Issue #36, opened 2023-09-01
verify.groovy:30-32: Implement XMIR generation. XMIR...


0pdd:
The puzzle `34-e7563e1a` from #34 has to be resolved: 

https://github.com/objectionary/jeo-maven-plugin/blob/22f719ec6be1c4dc9fa39d78e20b11c6fcf366af/src/it/successful-path/verify.groovy#L30-L32

The puzzle was created by @volodya-lombrozo on 01-Sep-23. 

Estimate: 90 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
The puzzle `34-e7563e1a` has disappeared from the source code, that's why I closed this issue.

0pdd:
@0pdd 2 puzzles [#39](https://github.com/objectionary/jeo-maven-plugin/issues/39), [#41](https://github.com/objectionary/jeo-maven-plugin/issues/41) are still not solved.

0pdd:
@0pdd 3 puzzles [#48](https://github.com/objectionary/jeo-maven-plugin/issues/48), [#49](https://github.com/objectionary/jeo-maven-plugin/issues/49), [#50](https://github.com/objectionary/jeo-maven-plugin/issues/50) are still not solved; solved: [#39](https://github.com/objectionary/jeo-maven-plugin/issues/39), [#41](https://github.com/objectionary/jeo-maven-plugin/issues/41).

0pdd:
@0pdd 2 puzzles [#48](https://github.com/objectionary/jeo-maven-plugin/issues/48), [#49](https://github.com/objectionary/jeo-maven-plugin/issues/49) are still not solved; solved: [#39](https://github.com/objectionary/jeo-maven-plugin/issues/39), [#41](https://github.com/objectionary/jeo-maven-plugin/issues/41), [#50](https://github.com/objectionary/jeo-maven-plugin/issues/50).

0pdd:
[...]

PR #38, opened 2023-09-01
feat(#36): Parse bytecode and save it into XMIR

volodya-lombrozo:
Parse bytecode and save it into XMIR. XMIR.xsd file was added in order to ensure XMIR compatibility with other tools.

Closes:#36


volodya-lombrozo:
@rultor merge

Issue #31, opened 2023-08-17
BytecodeIr.java:48-53: Continue implementing BytecodeIR...


0pdd:
The puzzle `26-78d4854d` from #26 has to be resolved: 

https://github.com/objectionary/jeo-maven-plugin/blob/08a22e3196487061ccaf408c2b46a329ee12cbb1/src/main/java/org/eolang/jeo/BytecodeIr.java#L48-L53

The puzzle was created by @volodya-lombrozo on 17-Aug-23. 

Estimate: 90 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

PR #61, opened 2023-09-08
feat(#31): Close deprecated puzzle

volodya-lombrozo:
Back transformation has already implemented. So the puzzle isn't relevant anymore.

Closes: #31


volodya-lombrozo:
@rultor merge

Issue #211, opened 2023-10-30
Distilled IT Prints Exceptions


volodya-lombrozo:
Although `distilled` integration test passes, it still has some issues. During the execution it throws the next exception:
```
org.objectweb.asm.tree.analysis.AnalyzerException: Error at instruction 4: Type java.lang.ClassNotFoundException: org.eolang.jeo.A$B not present
	at org.objectweb.asm.tree.analysis.Analyzer.analyze(Analyzer.java:273)
	at org.objectweb.asm.util.CheckClassAdapter.verify(CheckClassAdapter.java:1074)
	at org.objectweb.asm.util.CheckClassAdapter.verify(CheckClassAdapter.java:1032)
	at org.eolang.jeo.representation.bytecode.BytecodeClass.bytecode(BytecodeClass.java:151)
	at org.eolang.jeo.representation.xmir.XmlBytecode.bytecode(XmlBytecode.java:79)
	at org.eolang.jeo.representation.EoRepresentation.toBytecode(EoRepresentation.java:67)
	at org.eolang.jeo.improvement.ImprovementBytecodeFootprint.recompile(ImprovementBytecodeFootprint.java:71)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at org.eolang.jeo.improvement.ImprovementBytecodeFootprint.apply(ImprovementBytecodeFootprint.java:60)
	at org.eolang.jeo.improvement.ImprovementSet.apply(ImprovementSet.java:71)
	at org.eolang.jeo.Optimization.apply(Optimization.java:62)
	at org.eolang.jeo.JeoMojo.execute(JeoMojo.java:78)
	at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo(DefaultBuildPluginManager.java:126)
	at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute2(MojoExecutor.java:328)
	at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute(MojoExecutor.java:316)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:212)
	at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:174)
	at org.apache.maven.lifecycle.internal.MojoExecutor.access$000(MojoExecutor.java:75)
	at org.apache.maven.lifecycle.internal.MojoExecutor$1.run(MojoExecutor.java:162)
[...]

PR #212, opened 2023-10-30
feat(#211): Fix Bug With Broken Decorator

volodya-lombrozo:
Original problem wasn't related to bytecode check. Actually bytecode check alarmed about broken decorator class after transformation.
So, I fixed the bug and the problem is gone.

Closes: #211.
____
History:
- feat(#211): use ugly classloader to check bytecode correctly
- feat(#211): ignore original decorator
- feat(#211): fix all qulice suggestions

---

## PR-Codex overview
This PR focuses on improving the generation of distilled objects. 

### Detailed summary
- Added an assertion to check that a specific error message is not present in the log.
- Renamed the variable `additional` to `generated`.
- Updated the log message to include the names and count of the generated classes.
- Replaced the usage of `additional` with `generated` in the return statement.
- Added a condition to skip a specific decorator during object replacement.

> ✨ Ask PR-Codex anything about this PR by commenting with `/codex {your question}`

### DIFF
--- a/src/it/distilled/verify.groovy
+++ b/src/it/distilled/verify.groovy
@@ -25,6 +25,7 @@
 String log = new File(basedir, 'build.log').text;
 assert log.contains("BUILD SUCCESS")
 assert log.contains("Result: 31")
+assert !log.contains("java.lang.ClassNotFoundException")
 //Check that we have generated all the required XMIR object files.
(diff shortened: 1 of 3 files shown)


volodya-lombrozo:
@rultor merge

PR #243, opened 2023-11-08
feat(#165): Handle Fields of a Decorator Object Correctly

volodya-lombrozo:
Previousely we just removed all the fields from a decorator object. Now we handle them correctly and remove only inline candidates.

Closes: #165.
____
History:
- feat(#165): add integration test changes that shows what we want to achieve
- feat(#165): leave decorator fields
- feat(#165): fild initialization in constructors don't work
- feat(#165): refactor the code a bit
- feat(#165): fix all qulice suggestions

---

### Focus of this PR:
This PR focuses on improving the inlining optimization in the code.

### Detailed summary:
- The `name()` method in `XmlClass` is changed to public.
- An `int` variable `x` is added in class `B`.
- The `bar()` method in class `B` is modified to assign a value to `x` and return the sum of `a.foo()` and `x`.
- The `removeFieldsThatWillBeInlined()` method in `ImprovementDistilledObjects` is added to remove fields that will be inlined during optimization.
- The `isInlined()` method is added to check if a field will be inlined.

> ✨ Ask PR-Codex anything about this PR by commenting with `/codex {your question}`

### DIFF
--- a/src/it/distilled/src/main/java/org/eolang/jeo/B.java
+++ b/src/it/distilled/src/main/java/org/eolang/jeo/B.java
@@ -1,13 +1,14 @@
 package org.eolang.jeo;
 
-public final class B{
+public final class B {
     private final A a;
+    private int x = 2;
 
-    B(A a){
+    B(A a) {
         this.a = a;
     }
 
-    int bar(){
-        return a.foo() + 2;
+    int bar() {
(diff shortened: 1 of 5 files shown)


volodya-lombrozo:
@rultor merge

Issue #324, opened 2023-12-08
Application.java:90-94: Add integration test with...


0pdd:
The puzzle `320-aaedabfa` from #320 has to be resolved: 

https://github.com/objectionary/jeo-maven-plugin/blob/4aab01dfcfb106ff5506353255d1ffe0e457c83c/src/it/exceptions/src/main/java/org/eolang/jeo/exceptions/Application.java#L90-L94

The puzzle was created by @rultor on 08-Dec-23. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

PR #326, opened 2023-12-11
feat(#324): add test case with suppressed exception

volodya-lombrozo:
Add integration test case that throws and prints suppressed exception.

Closes: #324.


volodya-lombrozo:
@rultor merge

Issue #390, opened 2023-12-28
Hide empty `trycatchblocks`


volodya-lombrozo:
Now we print empty `trycatchblocks` for all the method, which is weird. It would be better just to omit it entirely.
Now:
```eo
  [arg__[Ljava/lang/String;__0] > j$main
    9 > access
    "([Ljava/lang/String;)V" > descriptor
    "" > signature
    * > exceptions
    seq > @
      tuple
        java/lang/StringBuilder
        .new
          "abc"
        .append
          1
        opcode > RETURN
          177
    tuple > trycatchblocks
```
I want:
```eo
  [arg__[Ljava/lang/String;__0] > j$main
    9 > access
    "([Ljava/lang/String;)V" > descriptor
    "" > signature
    * > exceptions
    seq > @
      tuple
        java/lang/StringBuilder
        .new
          "abc"
        .append
          1
        opcode > RETURN
          177
```

PR #393, opened 2023-12-29
feat(#390): Hide trycatchblocks directive 

volodya-lombrozo:
Hide `trycatchblocks` directives from final XMIR if a method doesn't have any try-catch blocks.

Closes: #390.
____
History:
- feat(#390): add trycatchblocks only if not empty
- feat(#390): add unit test


volodya-lombrozo:
@rultor merge

Issue #883, opened 2024-11-20
Not all `jeo` objects are defined by `+alias`


yegor256:
I see this code in XMIR generated by `jeo:disassemble`:

```
  [...]
   <metas>
      <meta>
         <head>package</head>
         <tail>j$org.j$eolang.j$benchmark</tail>
         <part>j$org.j$eolang.j$benchmark</part>
      </meta>
      <meta>
         <head>alias</head>
         <tail>jeo.opcode</tail>
         <part>jeo.opcode</part>
      </meta>
      <meta>
         <head>alias</head>
         <tail>jeo.label</tail>
         <part>jeo.label</part>
      </meta>
   </metas>
   <objects>
      <o abstract="" name="j$Big">
         <o base="jeo.int" line="643600858" name="version"><!-- 52 -->
            <o base="org.eolang.bytes" data="bytes">00 00 00 00 00 00 00 34</o>
         </o>
  [...]
```

Here, the `jeo.int` in the `@base` attribute doesn't mean `Q.jeo.int`, but instead it means `j$Big.jeo.int`. In order to mean `Q.jeo.int` it must be mentioned in the alias, just like it's done with the `jeo.label`. All `jeo.*` objects must have their aliases.

yegor256:
@volodya-lombrozo please, check

0pdd:
@yegor256 the puzzle [#890](https://github.com/objectionary/jeo-maven-plugin/issues/890) is still not solved.

volodya-lombrozo:
@yegor256 This one should be solved in the  `0.6.20` version.

0pdd:
@yegor256 the only puzzle [#890](https://github.com/objectionary/jeo-maven-plugin/issues/890) is solved here.

PR #888, opened 2024-11-21
feat(#883): Add All The Required Aliases

volodya-lombrozo:
In this PR I added inefficient solution for adding all the required `alias` metas.
This solution uses `xpath` on the pre-build XML. It means that instead of generation XML once, we generate it 2 times,
which is rather suboptimal, of course. I left one more puzzle to optimize it in the future.

Related to #883.


volodya-lombrozo:
@rultor merge

Issue #884, opened 2024-11-20
Redundant `LINENUMBER` opcodes may be removed


yegor256:
This is the bytecode for my Java method:

```
  // access flags 0x1
  public plain()J
  @Lorg/openjdk/jmh/annotations/Benchmark;()
   L0
    LINENUMBER 62 L0
    LCONST_0
    LSTORE 1
   L1
    LINENUMBER 63 L1
    ICONST_0
    ISTORE 3
   L2
   FRAME APPEND [J I]
    ILOAD 3
    GETSTATIC org/eolang/benchmark/Big.VALUES : [Ljava/lang/Object;
    ARRAYLENGTH
    IF_ICMPGE L3
   L4
    LINENUMBER 64 L4
    GETSTATIC org/eolang/benchmark/Big.VALUES : [Ljava/lang/Object;
    ILOAD 3
    AALOAD
    CHECKCAST java/lang/String
    INVOKEVIRTUAL java/lang/String.trim ()Ljava/lang/String;
    ASTORE 4
   L5
    LINENUMBER 65 L5
    ALOAD 4
    INVOKEVIRTUAL java/lang/String.length ()I
    ICONST_4
    IF_ICMPEQ L6
   L7
    LINENUMBER 66 L7
    GOTO L8
   L6
    LINENUMBER 68 L6
   FRAME APPEND [java/lang/String]
    LLOAD 1
    ALOAD 4
    BIPUSH 16
    INVOKESTATIC java/lang/Long.parseLong (Ljava/lang/String;I)J
    LCONST_1
    LADD
    LADD
    LSTORE 1
   L8
    LINENUMBER 63 L8
   FRAME CHOP 1
    IINC 3 1
    GOTO L2
   L3
    LINENUMBER 70 L3
   FRAME CHOP 1
    LLOAD 1
    LRETURN
   L9
    LOCALVARIABLE str Ljava/lang/String; L5 L8 4
    LOCALVARIABLE idx I L2 L3 3
    LOCALVARIABLE this Lorg/eolang/benchmark/Big; L0 L9 0
    LOCALVARIABLE acc J L1 L9 1
    MAXSTACK = 6
    MAXLOCALS = 5

```

I see many `LINENUMBER` opcodes (actually, one per every Java line), which are not necessary and may easily be eliminated before translating this code to XMIR. This will simplify XMIR and increase its readability (and speed of processing).

volodya-lombrozo:
[...]

PR #891, opened 2024-11-21
feat(#884): Skip Debug Labels

volodya-lombrozo:
In this PR I reduced the amount of labels generated by `disassemble` goal. We just exclude all the labels that are attached to line numbers.
Also, to allow this labels, I added the following option, that can be used to include line numbers and corresponding labels:

```xml
<configuration>
  <mode>debug</mode>
</configuration>
```

Related to #884.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,20 @@ configuration to your `pom.xml` file:
 </build>
 ```
 
+### Include debug information
+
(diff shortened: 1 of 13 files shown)


volodya-lombrozo:
@rultor merge

Issue #898, opened 2024-11-22
The `@version` attribute is not set correction of the XMIR


yegor256:
I'm using jeo 0.6.20, and I'm getting this XMIR:

```
<?xml version="1.0" encoding="UTF-8"?>
<program dob="2024-11-22T17:25:13.741113259Z"
         ms="1732296314430"
         name="j$Big"
         revision="0.0.0"
         time="2024-11-22T17:25:13.741113259Z"
         version="0.0.0">
[...]
```

What these `version="0.0.0"` means? I expect to see `version="0.6.20"` there. This `@version` attribute is the version of the generator of this XMIR. If it's JEO, here we should see the version of JEO.

volodya-lombrozo:
@rultor release, tag is `0.6.21`, title is `0.6.21`

PR #899, opened 2024-11-23
feat(#898): Set Correct Jeo Version to XMIR Files

volodya-lombrozo:
In this PR I fixed the problem with `jeo-maven-plugin` version in XMIR files.
We used to set the default `0.0.0` version, but now we read the current version from `MANIFEST.MF`

Related to: #898.


volodya-lombrozo:
@rultor merge

Issue #917, opened 2024-11-29
it's better to use different atoms for different opcodes


yegor256:
Currently, jeo:disassemble generates this phi (via XMIR):

```
φ ↦ Φ.jeo.seq.of6(
  α0 ↦ ξ.getstatic-25D7,
  α1 ↦ ξ.invokestatic-25E3,
  α2 ↦ ξ.invokedynamic-2600,
  α3 ↦ ξ.invokeinterface-2622,
  α4 ↦ ξ.invokeinterface-2627,
  α5 ↦ ξ.lreturn-2629
),

invokestatic-25E3 ↦ Φ.jeo.opcode( ... )

invokedynamic-2600 ↦ Φ.jeo.opcode( ... )
```

It's rather inconvenient for further processing. Instead, this would be better:

```
φ ↦ Φ.jeo.seq.of6(
  α0 ↦ ξ.a25D7,
  α1 ↦ ξ.a25E3,
  α2 ↦ ξ.a2600,
  α3 ↦ ξ.a2622,
  α4 ↦ ξ.a2627,
  α5 ↦ ξ.a2629
),

a25E3 ↦ Φ.jeo.opcode.invokestatic( ... )

a2600 ↦ Φ.jeo.opcode.invokedynamic( ... )
```

Possible?

yegor256:
@volodya-lombrozo please, check

yegor256:
@volodya-lombrozo actually, the name of the attribute (`getstatic-25D7`) doesn't matter at all, it may stay as is. What matters is the name of the atom/object being used (`jeo.opcode.invokestatic`).

volodya-lombrozo:
@yegor256 I don't clearly understand these `phi` expressions and haven't found any documentation on it. Can you please explain what does this mean:

`α0 ↦ ξ.a25D7` - is it an object with `base=α0` and `name=a25D7`  or something else?
What is the difference between `Φ` and `ξ`?

As for the issue itself, simply saying, do you want instead of:

```xml
<o base="jeo.opcode" name="invokespecial-188">
...
</o>
```

have something like:

```xml
<o base="jeo.opcode.invokespecial" name="a188">
...
</o>
```
?

yegor256:
@volodya-lombrozo yes, exactly

volodya-lombrozo:
[...]

PR #921, opened 2024-11-29
feat(#917): Different Atoms For Different Opcodes

volodya-lombrozo:
In this PR I add opcode names to their base.
For example, we used to have:
```
jeo.opcode
```
Now we have
```
jeo.opcode.ireturn
```
And we did it for all the opcodes.

Related to #917.


volodya-lombrozo:
@rultor merge

Issue #965, opened 2025-01-15
JcabiXmlNode.java:161-163: Remove the ignore list from...


0pdd:
The puzzle `946-4162b501` from #946 has to be resolved: 

https://github.com/objectionary/jeo-maven-plugin/blob/471bd9cfd1e0f70aeebb4cd1967818286a4780b2/src/main/java/org/eolang/jeo/representation/xmir/JcabiXmlNode.java#L161-L163

The puzzle was created by rultor on 15-Jan-25. 

Estimate: 30 minutes,  role: DEV. 

If you have any technical questions, don't ask me, submit new tickets instead. The task will be \"done\" when the problem is fixed and the text of the puzzle is _removed_ from the source code. Here is more about [PDD](http://www.yegor256.com/2009/03/04/pdd.html) and [about me](http://www.yegor256.com/2017/04/05/pdd-in-action.html).

0pdd:
@0pdd the puzzle [#1008](https://github.com/objectionary/jeo-maven-plugin/issues/1008) is still not solved.

0pdd:
@0pdd the only puzzle [#1008](https://github.com/objectionary/jeo-maven-plugin/issues/1008) is solved here.

PR #969, opened 2025-01-16
feat(#965): Fix All XMIR Offences Related to the ’name-outside-of-abstract-object’ Check

volodya-lombrozo:
In this PR I changed the `name` attribute to `as` for all inner objects, in order to fix all the `xmir` offences related to the `name-outside-of-abstract-object`.

Related to #965.


volodya-lombrozo:
@rultor merge

PR #1014, opened 2025-02-26
feat(#974): Update `jsmith` up to `0.1.3`

volodya-lombrozo:
In this PR I updated jsmith library to the `0.1.3` version. Also I've found a bug in computation of max locals value from bytecode.

The last unused variable `missed` isn't counted:

```java
    public void ia() {
        long h$m$$;
        boolean U$qa = true, A = U$qa == false, missed;
    }
```

It isn't a bug actually. We just can not retrieve more info from bytecode method instructions . 

Related to #974.


volodya-lombrozo:
@rultor merge

Issue #1172, opened 2025-07-09
replace `base` with `@` attribute


yegor256:
Currently, we use this:

```
<o name="maxs">
  <o base="Q.org.eolang.string" name="base">
    <o base="Q.org.eolang.bytes">
      <o>51-2E-6A-65-6F-2E-6D-61-78-73</o>
    </o>
  </o>
  <o base="Q.org.eolang.number" name="val495">
    <o base="Q.org.eolang.bytes">
      <o>3F-F0-00-00-00-00-00-00</o>
    </o>
  </o>
  <o base="Q.org.eolang.number" name="val504">
    <o base="Q.org.eolang.bytes">
      <o>3F-F0-00-00-00-00-00-00</o>
    </o>
  </o>
</o>
```

I suggest we use this instead:

```
<o name="maxs">
  <o base="Q.jeo.maxs" name="@"/>  <!-- here! -->
  <o base="Q.org.eolang.number" name="val495">
    <o base="Q.org.eolang.bytes">
      <o>3F-F0-00-00-00-00-00-00</o>
    </o>
  </o>
  <o base="Q.org.eolang.number" name="val504">
    <o base="Q.org.eolang.bytes">
      <o>3F-F0-00-00-00-00-00-00</o>
    </o>
  </o>
</o>
```

This is more in line with EO object model.

yegor256:
@volodya-lombrozo this would be very helpful!

0crat:
@yegor256 🌟 Fantastic enhancement suggestion! You've earned +12 points. Keep an eye out for any [areas to improve](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) in the repository and feel free to [report bugs](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) using our [effective reporting guidelines](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html). Your current balance is [-368](https://www.eolang.org/zerocracy/objectionary-vitals.html). Every contribution counts! 💪

yegor256:
@volodya-lombrozo you can ignore this, it's safe

0crat:
[...]

PR #1190, opened 2025-07-21
feat(#1172): replace base attribute with @ in delegate objects

volodya-lombrozo:
Refactors EO object representation to use `@` attribute instead of nested `base` objects, aligning with EO object model standards.

Related to #1172


volodya-lombrozo:
@rultor merge

Issue #1282, opened 2025-09-02
`jeo` breaks something in kafka `clients:tests` task


maxonfjvipon:
This is the broken build: https://github.com/objectionary/kafka/actions/runs/17397021249/job/49381094837

Here is the config:
```
      group = 'build'
      description = 'Run hone after Java compile'
      sourceSets.main.output.classesDirs.files.each { dir ->
        commandLine 'mvn',
          "-Dhone.timeout=400",
          "-Dhone.threads=0",
          "-Dhone.skip-phino",
          "-Dhone.jeo-version=0.14.0",
          "-Dhone.grep-in=(66-69-6C-74-65-72|6D-61-70)",
          "-Dhone.target=${project.buildDir}",
          "-Dhone.classes=${project.buildDir.toPath().relativize(dir.toPath())}",
          '-Dhone.rules=streams/*',
          "org.eolang:hone-maven-plugin:${v}:pull",
          "org.eolang:hone-maven-plugin:${v}:optimize"
      }
    }
```

As you see, we use `--skip-phino` so `phino` is not involved, only `jeo` does `disassemble` + `assemble` and we get:
```
> Task :clients:shadowJar FAILED
ex
org.gradle.api.GradleException: Could not add file '/home/runner/work/kafka/kafka/clients/build/classes/java/main/org/apache/kafka/common/Endpoint.class' to ZIP '/home/runner/work/kafka/kafka/clients/build/libs/kafka-clients-4.0.0.jar'.
```

So code is compiled but jar can't be built. It seems jeo looses some information in `org/apache/kafka/common/Endpoint.class` but I'm not sure

maxonfjvipon:
@volodya-lombrozo FYI

volodya-lombrozo:
@maxonfjvipon Could you provide a bit more context, please? The bytecode without `jeo` intervention and bytecode after `jeo` disassemble/assemble.

maxonfjvipon:
> @maxonfjvipon Could you provide a bit more context, please? The bytecode without jeo intervention and bytecode after jeo disassemble/assemble.

@volodya-lombrozo unfortunately no, I didn't build it locally

maxonfjvipon:
@volodya-lombrozo I'll try build it locally now, but I'm not sure my machine will do it fast enough

0crat:
[...]

PR #1300, opened 2025-09-09
fix(#1282): prevent empty signatures from being added to variable table

volodya-lombrozo:
This PR fixes a bug in the `LocalVariable` handling.

```java
 * The problem lies in the 'signature' field of the LocalVariableNode,
 * if it is null, we save an empty string in XMIR and then try to decode it back into bytecode,
 * which leads to saving an empty signature in the variable table of the method,
 * instead of (expected) null.
```

Related to #1282

## Summary by CodeRabbit

- Bug Fixes
  - Resolved an issue where absent local variable signatures could be emitted as empty strings in bytecode. Such signatures are now correctly omitted, improving fidelity with original bytecode and compatibility with bytecode tooling.

- Tests
  - Added a regression test ensuring that round-tripping bytecode does not introduce empty local variable signatures, safeguarding against future regressions.


volodya-lombrozo:
@rultor merge

Issue #1330, opened 2025-09-16
Error while reading `InlineInfoAttribute` from `kafka.server.KafkaBroker`


maxonfjvipon:
This is the pipeline: https://github.com/objectionary/kafka/actions/runs/17765929441/job/50489213467

It seems jeo misses some `InlineInfoAttribute` attribute

volodya-lombrozo:
@rultor release, tag is `0.14.6`, title is `0.14.6`

volodya-lombrozo:
@maxonfjvipon can we close this one?

0crat:
@volodya-lombrozo 🎉 Great job closing this issue! You've earned 8 points, bringing your [Objectionary balance](https://www.eolang.org/zerocracy/objectionary-vitals.html) to +171. Don't forget to check your [Zerocracy balance](https://www.zerocracy.com/) too. Keep up the excellent work! 💪

PR #1331, opened 2025-09-16
fix(#1330): ignore ScalaInlineInfo attribute

volodya-lombrozo:
This PR enhances adds Scala 2 integration tests and excluses 'ScalaInlineInfo' attribute. This attribute relies on the constant pool structure (CP). When we reassemble the CP
we need to rebuild the 'ScalaInlineInfo', which is impossible in our case. By this reason, we just ingore it.

Related to #1330


volodya-lombrozo:
@rultor merge

PR #1471, opened 2026-04-24
perf(#1470): cache compiled PathMatchers in GlobFilter constructor

### LINKED ISSUE #1470: `GlobFilter.test()` recompiles glob patterns on every call
## What happens

```java
@Override
        .map(GlobFilter::matcher)
        .collect(Collectors.toSet());
        .map(GlobFilter::matcher)
        .collect(Collectors.toSet());
    ...
}
```

## What should happen

```java

        .map(GlobFilter::matcher)
        .collect(Collectors.toSet());
        .map(GlobFilter::matcher)
        .collect(Collectors.toSet());
}
```

idrisoffrinat-cpu:
Closes #1470.

## Problem

`GlobFilter.test()` recompiled every include/exclude glob into a `PathMatcher`
on every invocation, even though `includes` and `excludes` are `final` and
immutable. For a plugin run that filters thousands of `.class` files, each
pattern was compiled N times instead of once.

## Fix

Move the `PathMatcher` compilation to the constructor and store the resulting
sets as `final` fields. `test()` now reads precomputed `whitelist` / `blacklist`
instead of rebuilding them per call. The matcher-building logic itself is
unchanged (same `GlobFilter::matcher` helper, same `Collectors.toSet()`).

This also aligns with the immutable-object principle the class already follows
for its string fields — all state is established at construction time.

## Testing

`mvn test -Dtest=GlobFilterTest` — 12/12 green, no new or modified tests
needed since behaviour is identical.

```
[INFO] Running org.eolang.jeo.GlobFilterTest
[INFO] Tests run: 12, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```

## Release Notes

* **Performance Improvements**
  * Optimized glob pattern matching performance by precompiling patterns during initialization, reducing computational overhead on subsequent operations.

### DIFF
--- a/src/main/java/org/eolang/jeo/GlobFilter.java
+++ b/src/main/java/org/eolang/jeo/GlobFilter.java
@@ -33,6 +33,16 @@ public final class GlobFilter implements Predicate<Path> {
     private final Set<String> excludes;
 
     /**
+     * Compiled matchers for {@link #includes}.
+     */
+    private final Set<PathMatcher> whitelist;
+
+    /**
+     * Compiled matchers for {@link #excludes}.
+     */
+    private final Set<PathMatcher> blacklist;
+
+    /**
      * Ctor.
      *
      * @param includes Glob patterns to include
@@ -41,6 +51,12 @@ public final class GlobFilter implements Predicate<Path> {
     GlobFilter(final Set<String> includes, final Set<String> excludes) {
         this.includes = includes;
         this.excludes = excludes;
+        this.whitelist = includes.stream()
+            .map(GlobFilter::matcher)
+            .collect(Collectors.toSet());
+        this.blacklist = excludes.stream()
+            .map(GlobFilter::matcher)
+            .collect(Collectors.toSet());
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


src/main/java/org/eolang/jeo/GlobFilter.java
```diff
@@ -41,6 +51,12 @@ public final class GlobFilter implements Predicate<Path> {
     GlobFilter(final Set<String> includes, final Set<String> excludes) {
         this.includes = includes;
         this.excludes = excludes;
+        this.whitelist = includes.stream()
```
volodya-lombrozo, line 55: @idrisoffrinat-cpu Can we create a separate constructor for the GolbFilter that will accept all four params? `GlobFilter(final Set<String> includes, final Set<String> excludes, Set<PathMatcher> whitelist, private final Set<PathMatcher> blacklist;)`

volodya-lombrozo (changes requested):
@idrisoffrinat-cpu great thanks for the contribution, just a small suggestion.

volodya-lombrozo:
Closing this one since there is no more progress.

PR #1473, opened 2026-04-24
Cache compiled PathMatcher instances in GlobFilter (#1470)

### LINKED ISSUE #1470: `GlobFilter.test()` recompiles glob patterns on every call

bibonix:
Closes #1470.

## Summary

Move compilation of the `includes` and `excludes` glob patterns from `GlobFilter.test(Path)` into the constructor and store the compiled `PathMatcher` instances as `final` instance fields. `test(Path)` now uses the cached matchers directly, so the glob patterns are compiled exactly once per `GlobFilter` instead of once per filtered path.

## Reproduction test

Added two tests in `GlobFilterTest` that fail against `master` and pass with the fix:

- `compilesIncludePatternsAtConstruction` -- an invalid include glob (`{unclosed`) is rejected at construction time with `PatternSyntaxException`, not silently accepted and only thrown on the first call to `test()`.
- `compilesExcludePatternsAtConstruction` -- same contract for exclude globs (`[abc`).

Both tests are a direct consequence of compiling the matchers once at construction, which is the fix requested in the issue.

## Why this is a fix

`FileSystems.getDefault().getPathMatcher(...)` compiles the glob on every call. In a Maven plugin that filters hundreds or thousands of `.class` files, the old `test()` recompiled every pattern once per filtered path. Since `includes` and `excludes` are immutable, the matchers are fully determined by constructor arguments and belong in `final` fields.

## Summary by CodeRabbit

* **Refactor**
  * Optimized glob pattern compilation for filtering operations to improve performance by precompiling patterns during initialization rather than on each invocation.

* **Tests**
  * Added validation tests to ensure invalid glob patterns are detected during initialization, providing earlier error feedback.

### DIFF
--- a/src/main/java/org/eolang/jeo/GlobFilter.java
+++ b/src/main/java/org/eolang/jeo/GlobFilter.java
@@ -33,6 +33,16 @@ public final class GlobFilter implements Predicate<Path> {
     private final Set<String> excludes;
 
     /**
+     * Compiled include matchers.
+     */
+    private final Set<PathMatcher> whitelist;
+
+    /**
+     * Compiled exclude matchers.
+     */
+    private final Set<PathMatcher> blacklist;
+
+    /**
      * Ctor.
      *
      * @param includes Glob patterns to include
@@ -41,6 +51,12 @@ public final class GlobFilter implements Predicate<Path> {
     GlobFilter(final Set<String> includes, final Set<String> excludes) {
         this.includes = includes;
         this.excludes = excludes;
+        this.whitelist = includes.stream()
+            .map(GlobFilter::matcher)
+            .collect(Collectors.toSet());
+        this.blacklist = excludes.stream()
+            .map(GlobFilter::matcher)
+            .collect(Collectors.toSet());
     }
 
     @Override
@@ -74,17 +90,11 @@ public final class GlobFilter implements Predicate<Path> {
 
     @Override
     public boolean test(final Path path) {
-        final Set<PathMatcher> whitelist = this.includes.stream()
-            .map(GlobFilter::matcher)
-            .collect(Collectors.toSet());
-        final Set<PathMatcher> blacklist = this.excludes.stream()
-            .map(GlobFilter::matcher)
-            .collect(Collectors.toSet());
         final boolean included;
-        if (blacklist.stream().anyMatch(m -> m.matches(path))) {
+        if (this.blacklist.stream().anyMatch(m -> m.matches(path))) {
             included = false;
         } else {
-            included = whitelist.isEmpty() || whitelist.stream()
+            included = this.whitelist.isEmpty() || this.whitelist.stream()
                 .anyMatch(matcher -> matcher.matches(path));
         }
         return included;
--- a/src/test/java/org/eolang/jeo/GlobFilterTest.java
+++ b/src/test/java/org/eolang/jeo/GlobFilterTest.java
@@ -7,10 +7,13 @@ package org.eolang.jeo;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Set;
+import java.util.regex.PatternSyntaxException;
 import java.util.stream.Stream;
 import org.hamcrest.MatcherAssert;
 import org.hamcrest.Matchers;
+import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
 import org.junit.jupiter.params.ParameterizedTest;
 import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.MethodSource;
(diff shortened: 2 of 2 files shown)


bibonix:
@yegor256 this PR is ready for your review.

## What
Fixes #1470. `GlobFilter.test(Path)` used to recompile every glob pattern via `FileSystems.getDefault().getPathMatcher(...)` on every invocation. Since a Maven plugin calls `test()` once per filtered `.class` file, a run with N files and P patterns did N × P compilations instead of P.

## How
- Added two `final` instance fields on `GlobFilter`: `whitelist` (compiled includes) and `blacklist` (compiled excludes).
- Moved the compilation step from `test()` into the constructor, so each glob is compiled exactly once per `GlobFilter` instance.
- `test(Path)` now reads the cached matchers directly; all externally visible behavior is preserved.

## Tests (reproduction first, then fix)
Two new tests in `GlobFilterTest` pin the behavior down and fail against `master`:
- `compilesIncludePatternsAtConstruction` — an invalid include glob (`{unclosed`) is now rejected eagerly at construction with `PatternSyntaxException`, rather than silently accepted and only thrown the first time `test()` is called.
- `compilesExcludePatternsAtConstruction` — the same contract for exclude globs (`[abc`).

Both tests are a direct, behavioral consequence of compiling the matchers once at construction. The existing parameterized `GlobFilter` cases continue to pass unchanged.

Commits are split so the failing reproduction comes first and the fix comes second:
- `test(#1470)`: add reproduction for GlobFilter matcher caching
- `fix(#1470)`: compile GlobFilter matchers once at construction

All 20 CI checks (maven test matrix, integration-test, qulice check, pdd, reuse, xcop, lint, benchmark) are green.

yegor256:
@volodya-lombrozo WDYT?

volodya-lombrozo:
@bibonix @yegor256 isn't a duplicate of https://github.com/objectionary/jeo-maven-plugin/pull/1471 ?

yegor256:
@volodya-lombrozo yes, but with a test

bibonix:
@yegor256 yes, #1473 and #1471 apply the same constructor-time cache change to `GlobFilter`. The added value in #1473 is `compilesIncludePatternsAtConstruction` and `compilesExcludePatternsAtConstruction` in `GlobFilterTest`, which pin the eager-compile behaviour — #1471 omits these. Either PR can be closed in favour of the other.

src/main/java/org/eolang/jeo/GlobFilter.java
```diff
@@ -32,6 +32,16 @@ public final class GlobFilter implements Predicate<Path> {
      */
     private final Set<String> excludes;
 
+    /**
+     * Compiled include matchers.
+     */
+    private final Set<PathMatcher> whitelist;
```
volodya-lombrozo, line 38: @bibonix Can we use `catoos` primitives in this PR instead?

src/main/java/org/eolang/jeo/GlobFilter.java
```diff
@@ -41,6 +51,12 @@ public final class GlobFilter implements Predicate<Path> {
     GlobFilter(final Set<String> includes, final Set<String> excludes) {
         this.includes = includes;
         this.excludes = excludes;
+        this.whitelist = includes.stream()
```
volodya-lombrozo, line 54: @bibonix We need a separate constructor with 4 separate parameters.

src/main/java/org/eolang/jeo/GlobFilter.java
```diff
@@ -41,6 +51,12 @@ public final class GlobFilter implements Predicate<Path> {
...
+            .map(GlobFilter::matcher)
+            .collect(Collectors.toSet());
+        this.blacklist = excludes.stream()
```
volodya-lombrozo, line 57: @bibonix Can we avoid running any logic inside of the constructor? I guess, `cactoos` primates might help solve this issue by running this code once lazily when it's needed

volodya-lombrozo (changes requested):

yegor256:
This is superseded by #1580, which just got merged and fixes the same GlobFilter recompilation issue (lazily, preserving the current fail-on-first-use behavior for invalid globs rather than failing at construction). Not merging this one to avoid the conflict; closing this out is probably right but leaving that call to a maintainer.

---

yegor256:
Please forgive us for closing this pull request: it has merge conflicts with the current master branch and has been open for more than a month, so it does not meet our bar for staying open.

This is not a judgment on the contribution itself, just routine stale-PR cleanup. We would be glad if you resubmit a new pull request, perhaps similar to this one, but rebased on top of the latest master branch. Thank you for your effort and understanding.

PR #1522, opened 2026-06-02
#1470: cache compiled PathMatcher instances in GlobFilter

nacNAC333:
## Issue #1470 — GlobFilter recompiles patterns on every call

GlobFilter.compile() was being called on every test() invocation, recompiling PathMatcher instances unnecessarily. This causes significant performance degradation when scanning large directories.

### Fix
- Compile PathMatchers once in the constructor
- Cache compiled instances for reuse in test() calls
- Added null-safe handling for edge cases

### Verification
- All existing tests pass
- Performance improvement verified
- HoC: ~30 (within range)


nacNAC333:
Superseded by #1523 (correct branch naming)

PR #1523, opened 2026-06-03
#1470: cache compiled PathMatcher patterns in GlobFilter

nacNAC333:
Compiled glob patterns once in the constructor instead of
recompiling on every test() call. Stored as Set<PathMatcher>
and simplified the test() logic.

### DIFF
--- a/src/main/java/org/eolang/jeo/GlobFilter.java
+++ b/src/main/java/org/eolang/jeo/GlobFilter.java
@@ -24,24 +24,36 @@ import java.util.stream.Collectors;
 public final class GlobFilter implements Predicate<Path> {
 
     /**
-     * Includes glob patterns (compiled to PathMatchers).
+     * Includes glob patterns.
      */
-    private final Set<PathMatcher> includes;
+    private final Set<String> includes;
 
     /**
-     * Excludes glob patterns (compiled to PathMatchers).
+     * Excludes glob patterns.
      */
-    private final Set<PathMatcher> excludes;
+    private final Set<String> excludes;
+
+    /**
+     * Compiled include matchers.
+     */
+    private final Set<PathMatcher> compiled;
+
+    /**
+     * Compiled exclude matchers.
+     */
+    private final Set<PathMatcher> rejected;
 
     /**
      * Ctor.
      *
-     * @param patterns Glob patterns to include
-     * @param exclusions Glob patterns to exclude
+     * @param includes Glob patterns to include
+     * @param excludes Glob patterns to exclude
      */
-    GlobFilter(final Set<String> patterns, final Set<String> exclusions) {
-        this.includes = GlobFilter.compile(patterns);
-        this.excludes = GlobFilter.compile(exclusions);
+    GlobFilter(final Set<String> includes, final Set<String> excludes) {
+        this.includes = includes;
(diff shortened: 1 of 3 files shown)


src/main/java/org/eolang/jeo/GlobFilter.java
```diff
@@ -49,45 +50,44 @@ public String toString() {
...
             inclusions = "no inclusions";
         } else {
-            inclusions = this.includes.stream().collect(
-                Collectors.joining(
-                    ", ",
-                    String.format("%d inclusions (", this.includes.size()),
-                    ")"
-                )
+            inclusions = String.format(
+                "%d patterns (compiled)",
```
edmoffo: The previous toString rendered each pattern verbatim, e.g. `1 inclusions (**/*.java)`, so a maintainer reading a log could see which globs the filter actually held. Replacing the body with `%d patterns (compiled)` drops that information and forces the existing `convertsGlobFilterToString` cases to be loosened. Keep the cache, but store the source strings alongside the compiled matchers and render them as before, so the diagnostic value of toString is preserved.

src/main/java/org/eolang/jeo/GlobFilter.java
```diff
@@ -23,24 +24,24 @@
...
      *
-     * @param includes Glob patterns to include
-     * @param excludes Glob patterns to exclude
+     * @param patterns Glob patterns to include
+     * @param exclusions Glob patterns to exclude
      */
-    GlobFilter(final Set<String> includes, final Set<String> excludes) {
-        this.includes = includes;
-        this.excludes = excludes;
+    GlobFilter(final Set<String> patterns, final Set<String> exclusions) {
```
edmoffo: The constructor parameters were `includes` and `excludes`, matching the field names and the wording of issue #1470. Renaming them to `patterns` and `exclusions` is unrelated to caching PathMatchers and should be reverted to keep the diff scoped to the cache change.

edmoffo (commented):
Two scope concerns below: the toString output now hides the actual glob patterns the filter was constructed with, and the constructor parameters were renamed unrelated to the cache work. Verdict: COMMENT — two inline comments, no correctness defect; CI status was unreported on this branch at the time of review.

src/main/java/org/eolang/jeo/GlobFilter.java
nacNAC333: Both addressed — source strings are stored alongside compiled matchers so toString still shows actual glob patterns, and constructor params are back to includes/excludes.

src/main/java/org/eolang/jeo/GlobFilter.java
nacNAC333: Reverted — constructor params are back to includes/excludes.

volodya-lombrozo:
I'm closing this one since the pipelines fail and there is no progress anymore.

0crat:
@edmoffo Thanks for the review! You've earned +7 points for this: +12 as a basis; -5 for very few (2) comments.  Your running score is [+43](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1541, opened 2026-06-16
fix #1535 drop stray 'of' from Caching.transform() error

### LINKED ISSUE #1535: Typo in Caching.transform() error message: 'Failed to transform of'
The error message produced when a transformation fails reads `Failed to transform of '...' to '...'`, which is not grammatically correct English — the preposition `of` does not belong there. The string appears once in `src/main/java/org/eolang/jeo/Caching.java` at line 52, inside the `IllegalStateException` raised by `transform()` when the underlying `tryTransform()` throws an `IOException`.

Because this message is the only thing a user sees in the build log when caching writes fail, the broken wording makes the report harder to read and harder to grep for.

The fix is the smallest possible: drop the word `of` so the format string becomes `Failed to transform '%s' to '%s'`. No callers need to change.

kreinba:
The format string at src/main/java/org/eolang/jeo/Caching.java:52 carries an extra `of` between `transform` and the source path, producing the ungrammatical `Failed to transform of '<src>' to '<dst>'` inside the IllegalStateException that wraps an IOException from tryTransform(). Issue #1535 traces the wording to that single literal and asks for `of` to be dropped, leaving `Failed to transform '%s' to '%s'`.

The change removes one word from the format string. No callers move, no signatures change, and the only difference at runtime is the wording of the log line a user sees when a cached transformation fails to write through.

The edit is text-only inside a runtime error message, so no behaviour, control flow, or assertion is touched. A grep for the old phrasing returns no remaining occurrences, the new phrasing already matches the wording used by sibling class `Informative` at src/main/java/org/eolang/jeo/Informative.java:47, and no existing test fixture pins the old string. The full Maven build was not run locally inside this routine.

Closes #1535

### DIFF
--- a/src/main/java/org/eolang/jeo/Caching.java
+++ b/src/main/java/org/eolang/jeo/Caching.java
@@ -49,7 +49,7 @@ public final class Caching implements Transformation {
         } catch (final IOException exception) {
             throw new IllegalStateException(
                 String.format(
-                    "Failed to transform of '%s' to '%s'",
+                    "Failed to transform '%s' to '%s'",
                     this.source(),
                     this.target()
                 ),


0crat:
@kreinba Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (2) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+94](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1559, opened 2026-06-28
#1535 drop spurious 'of' in caching transform error message

### LINKED ISSUE #1535: Typo in Caching.transform() error message: 'Failed to transform of'

kreinba:
The error message produced by Caching.transform() when tryTransform() throws an IOException previously read "Failed to transform of '%s' to '%s'" — the preposition "of" does not belong in that position and made the build log message harder to read and to grep for.

The format string in src/main/java/org/eolang/jeo/Caching.java line 52 now reads "Failed to transform '%s' to '%s'", which is the only visible effect of the change. No callers are affected because the argument list is unchanged.

The change is a single-character edit confined to one string literal; the existing surrounding test coverage exercises the same path, and Caching.java compiles unchanged under the project's source set.

Closes #1535

---

### DIFF
--- a/src/main/java/org/eolang/jeo/Caching.java
+++ b/src/main/java/org/eolang/jeo/Caching.java
@@ -49,7 +49,7 @@ public final class Caching implements Transformation {
         } catch (final IOException exception) {
             throw new IllegalStateException(
                 String.format(
-                    "Failed to transform of '%s' to '%s'",
+                    "Failed to transform '%s' to '%s'",
                     this.source(),
                     this.target()
                 ),


volodya-lombrozo:
@rultor merge

0crat:
@kreinba Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (2) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+90](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1564, opened 2026-07-01
Replace α0 with data and as-bytes in generated XMIR

### LINKED ISSUE #1563: Replace "α0" attribute in Φ.bytes/Φ.string/Φ.number XMIR with data/as-bytes
JEO hard-codes the phi-calculus positional attribute `α0` in the XMIR it generates for data objects. It shows up in three places: the inner `<o>` data child of every `Q.org.eolang.bytes` object (`src/main/java/org/eolang/jeo/representation/directives/DirectivesBytes.java:72` and `:79`), the nested `bytes` object inside a `Q.org.eolang.string` (`src/main/java/org/eolang/jeo/representation/directives/DirectivesValue.java:258`), and the nested `bytes` object inside a `Q.org.eolang.number` (`src/main/java/org/eolang/jeo/representation/directives/DirectivesNumber.java:70`). A string, for instance, is rendered as:

```xml
<o base="Q.org.eolang.string">
  <o as="α0" base="Q.org.eolang.bytes">
    <o as="α0">48-65-6C-6C-6F</o>
  </o>
</o>
```

This is wrong because phi-calculus is dropping positional attributes such as `α0`, so any XMIR that leans on them will eventually stop parsing. The value of `as` should be a named binding instead: `data` for the bytes payload, and `as-bytes` for how a `bytes` object binds into its `string` or `number` parent. After the change no `α0` should remain anywhere in emitted XMIR.

The fix is to replace the three hard-coded `"α0"` literals. In `DirectivesBytes` change the inner `.attr("as", "α0")` on lines 72 and 79 to `"data"`. In `DirectivesNumber` (line 70) and in `DirectivesValue.eoObject` (line 258) change the `as` argument handed to `new DirectivesBytes(..., "α0")` to `"as-bytes"`.

yegor256:
Data objects used to carry the phi-calculus positional attribute `α0` in the XMIR that jeo generates. This drops it everywhere: the bytes payload now binds through `as="data"`, and a `bytes` object binds into its `string` or `number` parent through `as="as-bytes"`.

We're getting rid of positional attributes like `α0` in phi-calculus, so any XMIR that still leans on them will eventually stop parsing. Moving to named bindings now keeps the output valid ahead of that change.

The reader side already picks up the payload by reading the first child's text, so the round-trip is unaffected — the full representation test suite (907 tests) stays green. New unit tests cover bytes, strings, and numbers, and assert that `α0` no longer appears in the output.

Closes #1563

### DIFF
--- a/src/main/java/org/eolang/jeo/representation/directives/DirectivesBytes.java
+++ b/src/main/java/org/eolang/jeo/representation/directives/DirectivesBytes.java
@@ -69,14 +69,14 @@ public final class DirectivesBytes implements Iterable<Directive> {
             directives = new DirectivesClosedObject(
                 new EoFqn("bytes").fqn(),
                 this.as,
-                new Directives().add("o").attr("as", "α0").set(this.hex).up()
+                new Directives().add("o").attr("as", "data").set(this.hex).up()
             );
         } else {
             directives = new DirectivesClosedObject(
                 new EoFqn("bytes").fqn(),
                 this.as,
                 this.name,
-                new Directives().add("o").attr("as", "α0").set(this.hex).up()
+                new Directives().add("o").attr("as", "data").set(this.hex).up()
             );
         }
         return directives.iterator();
--- a/src/main/java/org/eolang/jeo/representation/directives/DirectivesNumber.java
+++ b/src/main/java/org/eolang/jeo/representation/directives/DirectivesNumber.java
@@ -67,7 +67,7 @@ final class DirectivesNumber implements Iterable<Directive> {
             new EoFqn("number").fqn(),
             this.as,
             this.name,
-            new DirectivesBytes(this.hex, "", "α0")
+            new DirectivesBytes(this.hex, "", "as-bytes")
         ).iterator();
     }
 }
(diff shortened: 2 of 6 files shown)


yegor256:
@volodya-lombrozo mind taking a look when you get a moment? Happy to clarify anything about the change if it is unclear.

volodya-lombrozo (approved):

volodya-lombrozo:
@rultor merge

0crat:
@yegor256 Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1425](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@volodya-lombrozo Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+512](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1573, opened 2026-07-11
fix(#1538): descriptive error instead of NPE in toString() and root()

AysajanE:
Closes #1538 (reported by @morphqdd; @volodya-lombrozo approved the change on Jul 6).

`BytecodeClasses.toString()` threw a bare `NullPointerException` and `root()` handed out `null` when the `input` directory was not set, while `classes()` already failed with a descriptive `IllegalStateException`. All three now go through one private `checked()` guard that throws the same descriptive message, so a misconfigured run fails with a clear diagnostic instead of an NPE at some later call site.

Tests first: two regression tests reproducing the bare NPE / silent null (red on `master`), then the fix turns them green. `mvn clean install -Pqulice` passes locally on JDK 11.

Disclosure: this change was implemented with AI assistance under my direct review; I read and own every line. A signed self-run record of the checks (contributor self-run, not maintainer verification): https://northset-oss.github.io/verification-pilot/

Issue #1538, opened 2026-06-16
`BytecodeClasses.toString()` and `root()` throw `NullPointerException` when `input` is null while `classes()` provides a descriptive error


morphqdd:
What happens: `BytecodeClasses.classes()` (line 84) guards against a null `input` field with `Objects.isNull(this.input)` and throws an `IllegalStateException` with a clear message. However, `toString()` (line 43) calls `this.input.toString()` and `root()` (line 53) returns `this.input` directly, both without any null check. Any caller that invokes `root()` or `toString()` on a `BytecodeClasses` instance constructed with a null path (e.g. when Maven does not configure the sources directory) will receive a bare `NullPointerException` with no diagnostic information, while the same situation inside `classes()` produces a helpful error message.

What should happen: `toString()` and `root()` should either apply the same null guard as `classes()` or, consistent with the "no null" principle, the constructor should reject null with `Objects.requireNonNull` so the error is caught at construction time with a clear message.

morphqdd:
@yegor256, please take a look when you have a moment.

morphqdd:
@yegor256, ping

yegor256:
@volodya-lombrozo WDYT?

volodya-lombrozo:
Yes, why not to add it. Please, go ahead

0crat:
@morphqdd Thanks for suggesting an enhancement! You've earned +12 points. If you find anything else in the repository that [doesn't look](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) as good as you might expect, [do not hesitate](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) to [report](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) bugs too. Your running score is [+708](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1579, opened 2026-08-13
fix(#1578): name() now fails with a descriptive error when the object name is absent

### LINKED ISSUE #1578: `XmirRepresentation#name` throws bare `NodeNotFoundException` instead of a descriptive error
## What happens

`XmirRepresentation#name` (`src/main/java/org/eolang/jeo/representation/XmirRepresentation.java`, lines 66-75) reads the root object's `name` attribute unguarded:

```java
public String name() {
    final XmlNode root = this.xml.root();
    return new ClassName(
        root.xpath("/object/metas/meta[head[text()]='package']/tail/text()")
            .stream()
            .findFirst()
            .orElse(""),
        root.xpath("/object/o/@name").get(0)
    ).full();
}
```

The sibling xpath three lines above (the package lookup) correctly handles a "no match" case with `.stream().findFirst().orElse("")`. The `name` xpath does not: it calls `.get(0)` directly on the result list.

If the top-level `<o>` element in an otherwise well-formed XMIR document has metas but no `name` attribute, `root.xpath("/object/o/@name")` returns an empty list, and `.get(0)` throws `com.jcabi.xml.ListWrapper$NodeNotFoundException` (which extends `IndexOutOfBoundsException`). The exception message dumps the raw XML and says `Index (0) is out of bounds (size=0)`, giving no indication of what actually went wrong (a missing `name` attribute).

Reproduced locally: built a standalone XMIR document with metas present but no `name` attribute on the top `<o>`, and called `new XmirRepresentation(xml).name()`:

```
THREW: com.jcabi.xml.ListWrapper$NodeNotFoundException: XPath '/object/o/@name' not found in '<object>...<o base="Q.org.eolang.number"/></object>': Index (0) is out of bounds (size=0)
```

This is the same bug class already fixed by the maintainers in this exact area of the codebase: PR #1573 (issue #1538) replaced a bare, undescriptive exception in the sibling class `BytecodeClasses` with a clear `IllegalStateException`. Several other closed issues (#939, #964, #967, #762, #803) fixed the same "bare technical exception instead of a descriptive one" problem elsewhere in this repo.

Note: `XmirRepresentation#name` currently has no callers within this repo's own `src/main` (only its own tests call it), so it is not on this repo's own critical runtime path today, but it is public API and may be used by dependent projects (`opeo-maven-plugin`, `hone-maven-plugin`).

## What should happen

`XmirRepresentation#name` should throw a descriptive exception when the root object has no `name` attribute, for example `IllegalStateException("Can't find the object name in XMIR from '<source>'")`, matching the fix pattern already applied to `BytecodeClasses` in PR #1573.

VasilevNStas:
Fixes #1578.

## Problem
`XmirRepresentation#name()` (`src/main/java/org/eolang/jeo/representation/XmirRepresentation.java`, lines 66-74) reads the top-level object's `name` attribute unguarded: `root.xpath("/object/o/@name").get(0)`. When the top-level `<o>` has metas but no `name` attribute, the xpath returns an empty list and `.get(0)` throws a bare `com.jcabi.xml.ListWrapper$NodeNotFoundException` (`Index (0) is out of bounds (size=0)`), with no hint about what actually went wrong.

The sibling xpath for the package lookup (three lines above) already handles the "no match" case with `.stream().findFirst().orElse("")`.

This is the same bug class already fixed by maintainers in this exact area: PR #1573 (issue #1538) replaced a bare, undescriptive exception in `BytecodeClasses` with a clear `IllegalStateException`.

## Solution
Guard the `name` xpath result before indexing: if the list is empty, throw a descriptive `IllegalStateException` including the XMIR source:

```java
throw new IllegalStateException(
    String.format("Can't find the object name in XMIR from '%s'", this.source)
);
```

## Tests
- `throwsDescriptiveExceptionWhenNameAttributeIsAbsent` in `XmirRepresentationTest` — fails on `master` (bare `NodeNotFoundException`) and passes on this branch.

Verified locally: `mvn clean install -Pqulice --errors` (960 tests, 0 failures).

### DIFF
--- a/src/main/java/org/eolang/jeo/representation/XmirRepresentation.java
+++ b/src/main/java/org/eolang/jeo/representation/XmirRepresentation.java
@@ -6,6 +6,7 @@ package org.eolang.jeo.representation;
 
 import com.jcabi.xml.XML;
 import java.nio.file.Path;
+import java.util.List;
 import org.eolang.jeo.representation.bytecode.Bytecode;
 import org.eolang.jeo.representation.xmir.JcabiXmlDoc;
 import org.eolang.jeo.representation.xmir.XmlDoc;
@@ -65,12 +66,18 @@ public final class XmirRepresentation {
      */
     public String name() {
         final XmlNode root = this.xml.root();
+        final List<String> names = root.xpath("/object/o/@name");
+        if (names.isEmpty()) {
+            throw new IllegalStateException(
+                String.format("Can't find the object name in XMIR from '%s'", this.source)
+            );
+        }
         return new ClassName(
             root.xpath("/object/metas/meta[head[text()]='package']/tail/text()")
                 .stream()
                 .findFirst()
                 .orElse(""),
-            root.xpath("/object/o/@name").get(0)
+            names.get(0)
         ).full();
     }
 
--- a/src/test/java/org/eolang/jeo/representation/XmirRepresentationTest.java
+++ b/src/test/java/org/eolang/jeo/representation/XmirRepresentationTest.java
@@ -34,6 +34,7 @@ import org.xembly.Xembler;
  *
  * @since 0.1.0
  */
+@SuppressWarnings("PMD.TooManyMethods")
 final class XmirRepresentationTest {
 
     /**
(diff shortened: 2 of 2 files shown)


yegor256:
Thanks for the contribution!

---

0crat:
@vasilevnstas Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (30) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+618](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1580, opened 2026-08-13
fix(#1539): compile glob matchers lazily and cache them with cactoos primitives

### LINKED ISSUE #1539: `GlobFilter.test()` rebuilds `PathMatcher` sets on every invocation
What happens: The `test()` method in `GlobFilter` (lines 77–82) creates `whitelist` and `blacklist` sets of `PathMatcher` objects on every call by streaming over the glob pattern strings. Since `GlobFilter` is used as a `Predicate<Path>` passed to `Stream.filter()` in `FilteredClasses.all()`, it is invoked once per class file in the project. For a project with hundreds of `.class` files and several include/exclude patterns, this means thousands of redundant `PathMatcher` constructions per build.

What should happen: The `PathMatcher` sets should be precomputed once in the `GlobFilter` constructor and stored as final fields, so `test()` reuses them without reallocation on each call.

VasilevNStas:
Fixes #1539.

## Problem
`GlobFilter.test()` (`src/main/java/org/eolang/jeo/GlobFilter.java`) rebuilds the `whitelist` and `blacklist` sets of `PathMatcher` on **every** invocation by streaming over the glob pattern strings. Since `GlobFilter` is a `Predicate<Path>` passed to `Stream.filter()` in `FilteredClasses.all()`, it runs once per `.class` file: for a project with hundreds of classes and several patterns this means thousands of redundant `PathMatcher` constructions per build.

## Solution
Follow the architecture decision from the discussion in #1539: use **cactoos primitives** (`Scalar`, `Sticky`, `Synced`) as suggested by @volodya-lombrozo — instead of eagerly precompiling matchers into `final` fields (as PR #1473 does).

- `whitelist`/`blacklist` are now `Scalar<Set<PathMatcher>>` built as `Synced<>(new Sticky<>(...))`, so each glob pattern is compiled **exactly once** (on the first `test()` call) and cached.
- `Synced` makes the cached computation thread-safe — important because `Disassembler` invokes the filter from a parallel stream (`ParallelTranslator`).
- An extra package-private constructor accepts a `Function<String, PathMatcher>` compiler to make the caching behavior testable.

Behavior is fully preserved (unlike #1473): glob compilation stays **lazy** — an invalid glob pattern still fails on the first `test()` call with `PatternSyntaxException`, not at construction time.

## Tests
- `compilesMatchersOnlyOnce` — injects a counting compiler and asserts each of the two patterns is compiled once across three `test()` calls (fails on `master` where compilation happens per call).
- `keepsGlobCompilationLazy` — guards against the eager-compilation regression: an invalid glob must not throw at construction, only on the first `test()`.

Verified locally: `mvn clean install -Pqulice --errors` (961 tests, 0 failures).

Note: this supersedes/contrasts with PR #1473 (same issue, eager `final`-field caching). Happy to close either one.

### DIFF
--- a/src/main/java/org/eolang/jeo/GlobFilter.java
+++ b/src/main/java/org/eolang/jeo/GlobFilter.java
@@ -8,8 +8,13 @@ import java.nio.file.FileSystems;
 import java.nio.file.Path;
 import java.nio.file.PathMatcher;
 import java.util.Set;
+import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
+import org.cactoos.Scalar;
+import org.cactoos.scalar.Sticky;
+import org.cactoos.scalar.Synced;
+import org.cactoos.scalar.Unchecked;
 
 /**
  * Filter that matches paths against a set of glob patterns.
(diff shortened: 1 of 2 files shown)


VasilevNStas:
@yegor256 plz review

yegor256:
Thanks for the contribution!

---

0crat:
@vasilevnstas Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+634](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1596, opened 2026-09-05
fix(#1460): remove eo-maven-plugin:print usage from integration tests

### LINKED ISSUE #1460: `eo-maven-plugin` requires Java 17
In most of the pipelines I can see the following error:

```
The plugin org.eolang:eo-maven-plugin:0.59.0 has unmet prerequisites: 
Error:      Required Java version 17 is not met by current version: 11.0.30
```

We either should up the Java version to 17, or to remove `eo-maven-plugin:print` usage.

VasilevNStas:
Fixes #1460.

## Problem
`eo-maven-plugin:print` (goal `print`) is used in two integration tests — `bytecode-to-eo` and `custom-transformations` — to convert the generated XMIR into EO. `eo-maven-plugin` requires Java 17, which conflicts with the plugin's Java 8 target and forces these ITs to be **skipped on JDK 11** (`invoker.java.version = 17+` in their `invoker.properties`).

The current workaround only skips the tests on older JDKs; it does not fix the root cause. This is why the issue got the `bug` label (added 2026-09-01).

## Key finding
The `print` output (`target/generated-sources/jeo-eo`) is **not asserted anywhere**:
- `bytecode-to-eo/verify.groovy` checks only the XMIR output (`<listing>`, comments, indentation).
- `custom-transformations/verify.groovy` has the EO assertion **commented out** (`//assert new File(...jeo-eo...).exists()`).

So `eo-maven-plugin:print` is effectively dead weight that only drags in the Java 17 requirement.

## Solution
Remove the `eo-maven-plugin:print` usage (this is the second option proposed in the issue body — "remove `eo-maven-plugin:print` usage"):
- `src/it/bytecode-to-eo/pom.xml` — drop the `Print EO` profile.
- `src/it/custom-transformations/pom.xml` — drop the `eo-maven-plugin` plugin block.
- Both `invoker.properties` — drop `invoker.java.version = 17+`, so the ITs run again on JDK 11 (more coverage, no Java 17 requirement in the default CI run).

XMIR validity in these ITs is still guarded by `xmirVerification` (objectionary/lints), so nothing is lost.

Verified locally: `mvn clean install -Pqulice --errors` (962 tests, 0 failures) — both ITs pass on local JDK 21 without the plugin; the default invoker run stays green.

### DIFF
--- a/src/it/bytecode-to-eo/invoker.properties
+++ b/src/it/bytecode-to-eo/invoker.properties
@@ -1,4 +1,3 @@
 # SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
 # SPDX-License-Identifier: MIT
 invoker.goals=clean process-classes
-invoker.java.version = 17+
--- a/src/it/bytecode-to-eo/pom.xml
+++ b/src/it/bytecode-to-eo/pom.xml
@@ -57,34 +57,4 @@
       </plugin>
     </plugins>
   </build>
-  <profiles>
-    <profile>
-      <id>Print EO</id>
-      <activation>
-        <jdk>[11,)</jdk>
-      </activation>
-      <build>
-        <plugins>
-          <plugin>
-            <groupId>org.eolang</groupId>
-            <artifactId>eo-maven-plugin</artifactId>
-            <version>0.63.0</version>
-            <executions>
-              <execution>
-                <id>convert-xmir-to-eo</id>
-                <phase>process-classes</phase>
-                <goals>
-                  <goal>print</goal>
-                </goals>
-                <configuration>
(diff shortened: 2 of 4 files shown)


VasilevNStas:
@yegor256 
plz review this

yegor256:
Thanks — good catch that the EO output wasn't even being asserted, so dropping the unused plugin and the JDK 17 requirement is a clean fix. Merged.

---

0crat:
@vasilevnstas Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+1282](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1634, opened 2026-09-07
#1470: compile the globs once, when the filter is built

### LINKED ISSUE #1470: `GlobFilter.test()` recompiles glob patterns on every call

morphqdd:
`test` compiled every include and every exclude pattern on each call, and the plugin calls it once
per class file, so a build over a few thousand classes compiled the same patterns a few thousand
times and threw the result away each time. The patterns come from two final sets and never change.

They are compiled when the filter is built now. The include and exclude strings stay as they were,
`toString` still reads them.

One thing changes for a caller: a broken glob is refused when the filter is made rather than when
it is first applied. The new test pins that down, since it is the only observable difference.

Closes #1470

### DIFF
--- a/src/main/java/org/eolang/jeo/GlobFilter.java
+++ b/src/main/java/org/eolang/jeo/GlobFilter.java
@@ -33,14 +33,49 @@ public final class GlobFilter implements Predicate<Path> {
     private final Set<String> excludes;
 
     /**
+     * Compiled include patterns.
+     */
+    private final Set<PathMatcher> whitelist;
+
+    /**
+     * Compiled exclude patterns.
+     */
+    private final Set<PathMatcher> blacklist;
+
+    /**
      * Ctor.
      *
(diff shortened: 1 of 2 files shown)


morphqdd:
This is already done on master. `86fbd40c5` ("fix(#1539): compile glob matchers lazily and cache them with cactoos primitives") wraps both sets in a cached `Scalar`, so `test` no longer compiles anything after the first call. Closing my pull request, the issue can be closed as well.

PR #1635, opened 2026-09-07
fix(#1608): quote XPath literals in DirectivesClassTest so the assertion is not vacuous

### LINKED ISSUE #1608: DirectivesClassTest uses an unquoted XPath literal, making the assertion always true
## What happens

`DirectivesClassTest.java:110` builds an XPath with an **unquoted** literal:

```java
String.format(
    "./o[contains(@name,DirectivesClassTest)]/o[contains(@name, 'name')]/o/o[text()='%s']",
    ...
)
```

`contains(@name,DirectivesClassTest)` has no quotes around `DirectivesClassTest`. The jcabi/XPath engine treats the bare token as an unknown function reference, which evaluates to an empty string, so the predicate becomes `contains(@name,'')` — **always true** for any element with a non-null `@name`.

As a result the test `appendsFullyQualifiedName` passes regardless of the actual class name in the XML. It silently asserts nothing — a false positive that will not catch a regression.

## Expected

Quote the literal: `contains(@name,'DirectivesClassTest')`, so the predicate actually verifies the fully-qualified name is present in the XMIR.

## Files
- `src/test/java/org/eolang/jeo/representation/directives/DirectivesClassTest.java:110`

VasilevNStas:
Fixes #1608.

## Problem
`DirectivesClassTest#appendsFullyQualifiedName` builds the XPath with an **unquoted** literal:

```java
String.format(
    "./o[contains(@name,DirectivesClassTest)]/o[contains(@name, 'name')]/o/o[text()='%s']",
    ...
)
```

The jcabi/XPath engine treats the bare token `DirectivesClassTest` as an unknown function reference,
evaluates it to an empty string, so the predicate becomes `contains(@name,'')` — **always true**.
The test passed regardless of the actual class name in the XML, silently asserting nothing.

## Solution / What
- Quoted the literal in the positive path: `contains(@name,'DirectivesClassTest')`, so the predicate
  verifies the fully-qualified name is actually present.
- Added a negative assertion proving the predicate is now sensitive: `contains(@name,'WrongClass')`
  must NOT match. This assertion fails on the old code (vacuous match) and passes after the fix.

## Changes
- `src/test/java/org/eolang/jeo/representation/directives/DirectivesClassTest.java` — quoted XPath
  literals + sensitivity guard; XML is now evaluated once into a local variable.

## Tests
- `DirectivesClassTest#appendsFullyQualifiedName` — reproduces the vacuity (fails on `master` with
  the negative assertion) and passes after quoting. Full `mvn clean install -Pqulice` is green
  (all tests + jacoco + qulice + jtcop).

### DIFF
--- a/src/test/java/org/eolang/jeo/representation/directives/DirectivesClassTest.java
+++ b/src/test/java/org/eolang/jeo/representation/directives/DirectivesClassTest.java
@@ -97,19 +97,29 @@ final class DirectivesClassTest {
     @Test
     void appendsFullyQualifiedName() throws ImpossibleModificationException {
         final String name = "org.eolang.jeo.representation.directives.DirectivesClassTest";
+        final String xml = new Xembler(
+            new DirectivesClass(
+                new ClassName(name),
+                new DirectivesClassProperties(name)
+            )
+        ).xml();
         MatcherAssert.assertThat(
             "Can't append fully qualified class name",
-            new Xembler(
-                new DirectivesClass(
-                    new ClassName(name),
-                    new DirectivesClassProperties(name)
-                )
-            ).xml(),
+            xml,
             XhtmlMatchers.hasXPath(
                 String.format(
                     "./o[contains(@name,DirectivesClassTest)]/o[contains(@name, 'name')]/o/o[text()='%s']",
                     new DirectivesValue(0, new Format(), name.replace('.', '/'))
                         .hex(new JavaCodec())
+                )
+            )
+        );
+        MatcherAssert.assertThat(
+            "Wrong class name must not match the class predicate",
+            xml,
+            Matchers.not(
+                XhtmlMatchers.hasXPath(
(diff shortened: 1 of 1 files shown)


VasilevNStas:
@yegor256 
plz review this

yegor256:
Thanks! Merged.

Good find: unquoted, `DirectivesClassTest` inside `contains()` is a location path, not a literal, so it resolved to an empty node-set, stringified to `""`, and `contains(x, "")` held for anything. The added negative assertion is what keeps it honest.

---

0crat:
@vasilevnstas Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (24) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2282](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1637, opened 2026-09-07
fix(#1629): combine method access modifiers with a bitwise OR instead of addition

### LINKED ISSUE #1629: BytecodeMethodProperties sums access flags with IntStream.sum() instead of bitwise OR
## What happens

`BytecodeMethodProperties.java:84` builds an `int` access-flags value from an array of modifiers:

```java
this(IntStream.of(modifiers).sum(), name, descriptor, signature, new String[0]);
```

`IntStream.sum()` **adds** the modifiers, but access flags are a **bit mask** — the correct combination is a bitwise OR. Adding works only when the flags have no overlapping bits (which happens to be true for JVM access flags: `PUBLIC=1`, `PRIVATE=2`, `PROTECTED=4`, `STATIC=8`, ...). It is a latent correctness trap: any future flag whose value shares bits with another (or a bit value used twice) would silently produce a wrong mask.

## Expected

Use a bitwise OR:

```java
this(
    Arrays.stream(modifiers).reduce(0, (a, b) -> a | b),
    name, descriptor, signature, new String[0]
);
```

## Files
- `src/main/java/org/eolang/jeo/representation/bytecode/BytecodeMethodProperties.java:84`

VasilevNStas:
Fixes #1629.

## Problem
`BytecodeMethodProperties.java:84` combined method access flags with `IntStream.of(modifiers).sum()`.
Access flags are a **bit mask**, so the correct combination is a bitwise OR. Addition produces a wrong
mask for any overlapping bit values (JVM's current flags are powers of two, so the bug is latent, but
it is a correctness trap: a duplicated or overlapping flag would silently produce a wrong mask).

## Solution / What
Replaced the additive combination with a bitwise OR reduction:

```java
IntStream.of(modifiers).reduce(0, (a, b) -> a | b)
```

## Changes
- `src/main/java/org/eolang/jeo/representation/bytecode/BytecodeMethodProperties.java` — `sum()` → `reduce(0, OR)`.
- `src/test/java/org/eolang/jeo/representation/bytecode/BytecodeMethodPropertiesTest.java` — new test class
  with `combinesAccessModifiersWithBitwiseOr`: overlapping modifiers `{3, 5}` must produce `3 | 5 == 7`,
  not `3 + 5 == 8`.

## Tests
- `BytecodeMethodPropertiesTest#combinesAccessModifiersWithBitwiseOr` fails on `master` (additive
  combination) and passes after the fix.
- Full `mvn clean install -Pqulice` is green (964 tests, qulice + jacoco + jtcop).


VasilevNStas:
@yegor256 
plz review

yegor256:
Merged. Thanks — a latent bug, but `sum()` on a bit mask is exactly the kind of thing worth fixing before it bites.

---

0crat:
@vasilevnstas Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (39) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2166](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1642, opened 2026-09-07
fix(#1613): drop sign-commits and make README version extraction stable in up.yml

### LINKED ISSUE #1613: up.yml uses sign-commits with the default GITHUB_TOKEN and a fragile version grep
## What happens

`.github/workflows/up.yml` has two problems:

1. **`sign-commits: true` with the default `GITHUB_TOKEN`** (line 28, `peter-evans/create-pull-request`). The default token is not backed by a GPG key configured in the runner, so commit signing cannot work — either the step fails or the commits end up unsigned, defeating the purpose of the flag.

2. **Fragile version extraction** (line 24):

```bash
prev=$(grep -oE '<version>[^<]+' README.md | sed 's/<version>//' | head -1)
```

`head -1` takes the **first** `<version>` occurrence in the README. Today that is the Maven Central badge version (line ~70), but any `<version>` added above it would silently break the auto-bump (the `sed` would then replace the wrong string in the whole file).

## Expected

- Drop `sign-commits: true` (or configure a signing key, e.g. `git config user.signingkey` + a private key secret).
- Extract the version from a specific, stable location (e.g. the badge URL line only), not the first `<version>` match.

## Files
- `.github/workflows/up.yml:24,28`

VasilevNStas:
Fixes #1613

## Problem
`.github/workflows/up.yml` had two issues:

1. `sign-commits: true` was set on `peter-evans/create-pull-request` while using the default
   `GITHUB_TOKEN`, which is not backed by a signing key — commit signing cannot work
2. Version extraction took the **first** `<version>` occurrence in `README.md`:
   ```bash
   prev=$(grep -oE '<version>[^<]+' README.md | sed 's/<version>//' | head -1)
   ```
   Any `<version>` added above the plugin snippet would silently break the auto-bump

## Solution / What
1. Removed `sign-commits: true`
2. Extracted the version from the plugin block only — the line right after
   `<artifactId>jeo-maven-plugin</artifactId>`:
   ```bash
   prev=$(awk '/<artifactId>jeo-maven-plugin<\/artifactId>/{getline; print}' README.md | grep -oE '<version>[^<]+' | sed 's/<version>//')
   ```

## Changes
- `.github/workflows/up.yml` — dropped `sign-commits`, replaced the fragile `grep | head -1`
  extraction with an `awk`-based one anchored to the plugin artifactId

## Tests
- Verified the new extraction returns `0.15.2` on the current `README.md`
- Verified robustness with a synthetic input where a foreign `<version>` appears above the plugin block —
  the plugin version is still extracted correctly
- `yamllint .github/workflows/up.yml` passes

### DIFF
--- a/.github/workflows/up.yml
+++ b/.github/workflows/up.yml
@@ -21,11 +21,10 @@ jobs:
       - run: |-
           git fetch --tags --force && \
           latest=$(git tag --sort=creatordate | tail -1) && \
-          prev=$(grep -oE '<version>[^<]+' README.md | sed 's/<version>//' | head -1) && \
+          prev=$(awk '/<artifactId>jeo-maven-plugin<\/artifactId>/{getline; print}' README.md | grep -oE '<version>[^<]+' | sed 's/<version>//') && \
           sed -i "s/${prev}/${latest}/g" README.md
       - uses: peter-evans/create-pull-request@v8
         with:
-          sign-commits: true
           branch: version-up
           commit-message: 'chore: new version in README'
           delete-branch: true


VasilevNStas:
@yegor256 
please review this

yegor256:
Merged, thanks! Anchoring the extraction to the artifactId is much better than `head -1`.

---

0crat:
@vasilevnstas Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (3) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2306](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1644, opened 2026-09-07
fix(#1616): restore the context classloader in PluginStartupTest and assert the loaded class is instantiable

### LINKED ISSUE #1616: PluginStartupTest leaves a dynamic classloader attached to the test thread
## What happens

`PluginStartupTest.loadsClassesDynamically` (`src/test/java/org/eolang/jeo/PluginStartupTest.java:27-39`) calls `new PluginStartup(project, dir).init()`, which replaces the **context classloader of the current thread** (`PluginStartup.init()` -> `Thread.currentThread().setContextClassLoader(new JeoClassLoader(...))`). The test never restores the original classloader, so:

- the classloader (and the dynamically loaded `SomeClassCompiledDynamically` class) leaks into any test that runs afterwards and reads the context classloader;
- the assertion only checks `notNullValue()` on the loaded `Class` object, not that the class is actually usable.

## Expected

Restore the original context classloader in a `finally` block (and ideally assert the loaded class is functional, e.g. via reflection `newInstance()`).

## Files
- `src/test/java/org/eolang/jeo/PluginStartupTest.java:27-39`
- `src/main/java/org/eolang/jeo/PluginStartup.java` (`init()`)

VasilevNStas:
Fixes #1616

## Problem
`PluginStartupTest#loadsClassesDynamically` called `new PluginStartup(...).init()`, which replaces the
**context classloader of the current thread** with a `JeoClassLoader`, and never restored it. The
classloader (and the dynamically loaded `SomeClassCompiledDynamically` class) leaked into any later test
that reads the context classloader. The assertion only checked `notNullValue()` on the loaded `Class`,
not that the class was actually usable

## Solution / What
- Restore the original context classloader in a `finally` block
- Build the test class with a real default constructor (`ALOAD 0; INVOKESPECIAL Object.<init>; RETURN`)
  and assert the loaded class is instantiable via `getDeclaredConstructor().newInstance()`
- Assert after the try/finally that the thread's context classloader equals the original (guards the leak)

## Changes
- `src/test/java/org/eolang/jeo/PluginStartupTest.java` — added constructor bytecode to the test class,
  try/finally restore, and two stronger assertions (instantiable + no leak)

## Tests
- Reproduced on `master`: the "must not leak" assertion fails (context classloader is `JeoClassLoader`)
- Passes after the fix; full `mvn clean install -Pqulice` green (963 tests, qulice + jacoco + jtcop)

### DIFF
--- a/src/test/java/org/eolang/jeo/PluginStartupTest.java
+++ b/src/test/java/org/eolang/jeo/PluginStartupTest.java
@@ -13,6 +13,7 @@ import org.hamcrest.MatcherAssert;
 import org.hamcrest.Matchers;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.io.TempDir;
+import org.objectweb.asm.Opcodes;
 
 /**
  * Test cases for {@link PluginStartup}.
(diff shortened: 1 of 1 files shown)


VasilevNStas:
@yegor256 
plz review this

zaguzovmaksim0-hue (approved):
I checked the test lifecycle and the generated class path rather than only the final assertion.

`PluginStartup.init()` mutates the current thread's context classloader, so saving it before initialization and restoring it in `finally` is the correct isolation boundary even if class loading or instantiation fails. The final equality assertion then verifies that isolation explicitly rather than relying on teardown order.

The stronger class assertion is also meaningful: the generated bytecode now has a real public constructor (`ALOAD 0`, `Object.<init>`, `RETURN`), and the test resolves the class through the classloader installed by `PluginStartup` and instantiates it before restoration. That proves both loading and basic usability while the intended loader is active.

The change is test-only, the full CI matrix is green, and I found no leak path left in this test. Approving.

yegor256:
The classloader restore is right, but the second assertion cannot fail: `original` is put back in the `finally` block immediately before it, so `getContextClassLoader()` is always equal to it at that point. A test that cannot fail is worse than no test. Assigning to @yegor256.

---

0crat:
@zaguzovmaksim0-hue Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (29) hits-of-code; +4 to give you at least something.  Your running score is [+52](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1645, opened 2026-09-07
test(#1615): assert only the stable error count instead of the full XSD message

### LINKED ISSUE #1615: XmirFilesTest asserts on the exact XSD validator message, which breaks on dependency upgrades
## What happens

`XmirFilesTest.java:174` asserts the **exact** text of the XML/XSD validation error:

```java
Matchers.containsString(
    "1 error(s) in XML document: cvc-complex-type.2.4.a: Invalid content was found starting with element 'tail'. One of '{head}' is expected."
)
```

This message comes from the underlying XML validator (jcabi/XML + the XSD schema). Its wording is not part of the plugin's own code and can change with a dependency upgrade (new jcabi/xml, a newer XML schema, or a different parser) — at which point the test fails without any real regression in the plugin.

## Expected

Assert only the *fact* of the failure (e.g. `IllegalArgumentException`), or match a stable fragment (like the error count), instead of the full vendor-specific message.

## Files
- `src/test/java/org/eolang/jeo/XmirFilesTest.java:174`

VasilevNStas:
Fixes #1615

## Problem
`XmirFilesTest.java:173-175` asserted the exact, vendor-specific XSD validation message:
`1 error(s) in XML document: cvc-complex-type.2.4.a: Invalid content was found starting with
element 'tail'. One of '{head}' is expected.` The wording comes from the underlying XML validator
(jcabi/XML + XSD schema), not from the plugin, so a dependency upgrade could break the test without
any real regression

## Solution / What
Assert only the stable error-count fragment (`1 error(s) in XML document`), which still verifies that
the schema validation runs and reports a single error for the invalid XMIR, while being resilient to
changes in the validator's message wording

## Changes
- `src/test/java/org/eolang/jeo/XmirFilesTest.java` — shortened the asserted fragment

## Tests
- `XmirFilesTest` passes (9 tests)

### DIFF
--- a/src/test/java/org/eolang/jeo/XmirFilesTest.java
+++ b/src/test/java/org/eolang/jeo/XmirFilesTest.java
@@ -171,7 +171,7 @@ final class XmirFilesTest {
                 () -> new XmirFiles(temp).verify()
             ).getMessage(),
             Matchers.containsString(
-                "1 error(s) in XML document: cvc-complex-type.2.4.a: Invalid content was found starting with element 'tail'. One of '{head}' is expected."
+                "1 error(s) in XML document"
             )
         );
     }


VasilevNStas:
@yegor256 
please review

zaguzovmaksim0-hue (approved):
This keeps the useful invariant and removes only the unstable one. The test still executes `XmirFiles.verify()`, still requires an `IllegalArgumentException`, and still checks that validation reports exactly one XML error. What it stops pinning is Xerces/jcabi wording for the specific schema violation, which is outside this plugin's contract and can change with dependency upgrades.

The change is one assertion fragment, the full CI matrix is green, and it does not weaken the test into merely "some exception happened". No correctness issue found. Approving.

0crat:
@zaguzovmaksim0-hue Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (2) hits-of-code; +4 to give you at least something.  Your running score is [+48](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1646, opened 2026-09-07
fix(#1628): remove the redundant global JAXP TransformerFactory registration

### LINKED ISSUE #1628: Saxon is force-registered as the global JAXP TransformerFactory though no JAXP transform is used
## What happens

`src/main/resources/META-INF/services/javax.xml.transform.TransformerFactory` contains `net.sf.saxon.TransformerFactoryImpl`, so the plugin registers Saxon as the **global JAXP `TransformerFactory`** for the whole classpath.

But the main code does not use JAXP transformers at all (it uses `org.xembly.Transformers` and eo-parser's `StrictXmir`). Registering Saxon globally affects every other JAXP consumer in the same classloader, which can change their behavior unexpectedly (e.g. a different XSLT 2.0 engine instead of the JDK default).

## Expected

Verify whether the registration is actually needed; if not, remove the `META-INF/services` file (or narrow its scope). If it is needed, document why.

## Files
- `src/main/resources/META-INF/services/javax.xml.transform.TransformerFactory`

VasilevNStas:
Fixes #1628

## Problem
`src/main/resources/META-INF/services/javax.xml.transform.TransformerFactory` force-registered Saxon
as the **global JAXP `TransformerFactory`** for the whole classpath. The plugin's own code does not use
JAXP transformers directly (it uses `org.xembly.Transformers` and eo-parser's `StrictXmir`), so the
registration only affected unrelated JAXP consumers in the same classloader

## Solution / What
Removed the service file. It was redundant: both **Saxon-HE 12.9** and **eo-parser 0.63.3** already ship
`META-INF/services/javax.xml.transform.TransformerFactory` = `net.sf.saxon.TransformerFactoryImpl`
(verified in their jars), and Saxon-HE is on the runtime classpath both directly and transitively via
eo-parser. JAXP `TransformerFactory.newInstance()` still resolves to Saxon

## Changes
- Removed `src/main/resources/META-INF/services/javax.xml.transform.TransformerFactory`

## Tests
- Verified the same registration exists in `Saxon-HE-12.9.jar` and `eo-parser-0.63.0.jar`
- Full `mvn clean install -Pqulice` green (all tests + qulice + jacoco + jtcop, no dependency problems)

### DIFF
--- a/src/main/resources/META-INF/services/javax.xml.transform.TransformerFactory
+++ b/src/main/resources/META-INF/services/javax.xml.transform.TransformerFactory
@@ -1,1 +1,0 @@
-net.sf.saxon.TransformerFactoryImpl


VasilevNStas:
@yegor256 
plz review

yegor256:
Merged, thanks! Checking that both Saxon-HE and eo-parser already ship the same registration is what made this safe to drop.

---

0crat:
@vasilevnstas Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (1) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2318](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1648, opened 2026-09-07
fix(#1601): compute the LDC stack impact from the constant kind without NPE

### LINKED ISSUE #1601: impact() in BytecodeInstruction computes wrong stack size for LDC with non-numeric constants
## What happens

`BytecodeInstruction.impact()` computes the stack effect of an instruction, and for `LDC` it derives the size from the *runtime class* of the constant (`src/main/java/org/eolang/jeo/representation/bytecode/BytecodeInstruction.java`, lines 276-285):

```java
case LDC: {
    final Class<?> clazz = this.args.get(0).getClass();
    if (clazz == Long.class || clazz == Double.class) {
        result = 2;
        break;
    } else {
        result = BytecodeInstruction.size(Type.getType(clazz));
        break;
    }
}
```

This is wrong for the constants that `LDC` can actually hold:

- a `null` literal -> `this.args.get(0)` is `null` -> `NullPointerException`;
- a `Type` constant (used for class literals, `LDC Type`) -> `Type.getType(Class)` throws `IllegalArgumentException` (ASM `Type` is not a `java.lang.Class` but `org.objectweb.asm.Type`);
- a `Handle` constant (`LDC Handle`, Java 7+) -> `Type.getType(Handle)` fails similarly;
- an array/`Class` constant -> size is guessed as 1, which is wrong for `long`/`double` arrays.

`impact()` feeds `MaxStack`/`MaxLocals` via `InstructionsFlow`, so a wrong value silently corrupts the computed `max_stack`/`max_locals` (or crashes) during `assemble`.

The same pattern affects `GETSTATIC`/`PUTSTATIC`/`GETFIELD`/`PUTFIELD` (lines 286-297), which do `Type.getType(String.valueOf(this.args.get(2)))` — the argument there is already a descriptor `String`, so it works, but it is fragile.

There is no direct unit test for `impact()` at all (`BytecodeInstructionTest` has a single test that never calls it).

## Expected

`impact()` should compute the correct stack size for every `LDC` operand (`Long`/`Double` -> 2, other primitives -> 1, `String` -> 1, `Type`/`Handle` -> 1, `null` -> 1 as `aconst_null`/`LDC null` pushes a reference), and be covered by unit tests for each operand kind.

## Files
- `src/main/java/org/eolang/jeo/representation/bytecode/BytecodeInstruction.java:276-285` (and 286-297)
- `src/test/java/org/eolang/jeo/representation/bytecode/BytecodeInstructionTest.java:23-44` — no coverage of `impact()`

VasilevNStas:
Fixes #1601.

## Problem
`BytecodeInstruction.impact()` computed the `LDC` stack effect from the **runtime class** of the
constant (`BytecodeInstruction.java:276-285`):
```java
final Class<?> clazz = this.args.get(0).getClass();
if (clazz == Long.class || clazz == Double.class) {
    result = 2;
} else {
    result = BytecodeInstruction.size(Type.getType(clazz));
}
```
This crashes or misbehaves for operands `LDC` can actually hold:
- `null` literal → `NullPointerException` (`getClass()` on `null`);
- `Type` / `Handle` constants → size derived from the wrong runtime class (works only by luck);
- `ConstantDynamic` → not handled.

`impact()` feeds `MaxStack`/`MaxLocals` via `InstructionsFlow`, so a wrong value silently corrupts
the computed `max_stack`/`max_locals` (or crashes) during `assemble`. There was no direct unit test
for `impact()`.

## Solution / What
Derive the size from the JVM semantics of `LDC`: only `Long`/`Double` constants push two slots;
every other constant (`Integer`, `Float`, `String`, `Type`, `Handle`, `ConstantDynamic`, `null`)
pushes one reference slot:
```java
final Object arg = this.args.get(0);
if (arg instanceof Long || arg instanceof Double) {
    result = 2;
} else {
    result = 1;
}
```

## Changes
- `src/main/java/org/eolang/jeo/representation/bytecode/BytecodeInstruction.java` — fixed the `LDC` case.
- `src/test/java/org/eolang/jeo/representation/bytecode/BytecodeInstructionTest.java` — added 8 tests
  for `impact()` covering `Long`, `Double`, `Integer`, `Float`, `String`, `Type`, `Handle`, and `null`.

## Tests
- `calculatesImpactForLdcNull` fails on `master` with `NullPointerException` and passes after the fix.
- Full `mvn clean install -Pqulice` green (971 tests, qulice + jacoco + jtcop).

### DIFF
--- a/src/main/java/org/eolang/jeo/representation/bytecode/BytecodeInstruction.java
+++ b/src/main/java/org/eolang/jeo/representation/bytecode/BytecodeInstruction.java
@@ -274,14 +274,13 @@ public final class BytecodeInstruction implements BytecodeEntry {
                 result = 2;
                 break;
             case LDC: {
-                final Class<?> clazz = this.args.get(0).getClass();
-                if (clazz == Long.class || clazz == Double.class) {
+                final Object arg = this.args.get(0);
+                if (arg instanceof Long || arg instanceof Double) {
                     result = 2;
-                    break;
                 } else {
-                    result = BytecodeInstruction.size(Type.getType(clazz));
-                    break;
+                    result = 1;
                 }
+                break;
             }
             case GETSTATIC:
                 result = BytecodeInstruction.size(Type.getType(String.valueOf(this.args.get(2))));
--- a/src/test/java/org/eolang/jeo/representation/bytecode/BytecodeInstructionTest.java
+++ b/src/test/java/org/eolang/jeo/representation/bytecode/BytecodeInstructionTest.java
@@ -9,6 +9,7 @@ import org.eolang.jeo.representation.directives.Format;
 import org.hamcrest.MatcherAssert;
 import org.hamcrest.Matchers;
 import org.junit.jupiter.api.Test;
+import org.objectweb.asm.Handle;
 import org.objectweb.asm.Opcodes;
 import org.objectweb.asm.Type;
 import org.xembly.ImpossibleModificationException;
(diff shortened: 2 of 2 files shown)


VasilevNStas:
@yegor256 
plz review

yegor256:
Merged. Thanks — deriving the slot count from JVM semantics rather than the operand's runtime class is right, and `calculatesImpactForLdcNull` pins the crash.

---

0crat:
@vasilevnstas Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+2254](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

Issue #1620, opened 2026-09-06
Wire the six bundled XSL normalization stylesheets into the pipeline or remove them


VasilevNStas:
## What happens

The plugin ships six XSL stylesheets under `src/main/resources/org/eolang/parser/` (`add-cuts.xsl`, `add-refs.xsl`, `remove-cuts.xsl`, `roll-bases.xsl`, `vars-float-down.xsl`, `wrap-method-calls.xsl`) that are **never referenced by any Java code** (a grep over `src/**` finds no loader; the pipeline only uses `StrictXmir` from eo-parser). They are dead resources that:

- sit on the same classpath path (`org/eolang/parser/`) as the identically-named XSL inside `eo-parser`, so a future name-based load would silently shadow the original files;
- duplicate the normalization logic that eo-parser already applies.

See also issue #1602 for the duplicated `id="add-refs"` in `add-cuts.xsl`/`remove-cuts.xsl`.

## Expected

Either integrate these stylesheets into the XMIR normalization pipeline of `disassemble` (they were clearly meant for that: add-refs/add-cuts/remove-cuts/vars-float-down/roll-bases/wrap-method-calls), or remove them from the plugin resources.

## Files
- `src/main/resources/org/eolang/parser/*.xsl` (6 files)

VasilevNStas:
This issue is resolved by the removal of the six dead stylesheets in **PR #1649** (for #1602).

Decision: **remove** (the second option the issue offers). The investigation showed:

- No Java code references the six `org/eolang/parser/*.xsl` files (verified by `grep` over `src/**`).
- eo-parser already applies XMIR normalization through its own pipeline stylesheets under
  `org/eolang/parser/parse/` (used by `StrictXmir`), so the plugin copies were dead duplicates.
- The plugin copies sat at a different classpath path (`org/eolang/parser/*.xsl` vs eo-parser's
  `org/eolang/parser/parse/*.xsl`), so no shadowing occurs today — removing them eliminates the
  future hazard entirely (including the duplicated `id="add-refs"` noted in #1602).

`src/main/resources/org/eolang/parser/*.xsl` no longer exists on the `fix/issue-1602-xsl` branch.
Closing as resolved by #1649 (merge pending).

0crat:
[...]

PR #1649, opened 2026-09-07
fix(#1602): remove six dead XSL stylesheets with a duplicated id='add-refs'

### LINKED ISSUE #1602: add-cuts.xsl and remove-cuts.xsl both declare id='add-refs' and are never loaded
## What happens

Six XSL stylesheets under `src/main/resources/org/eolang/parser/` (`add-cuts.xsl`, `add-refs.xsl`, `remove-cuts.xsl`, `roll-bases.xsl`, `vars-float-down.xsl`, `wrap-method-calls.xsl`) are **never loaded by any Java code** — a `grep` over `src/**` finds no reference to them (the pipeline only uses `StrictXmir` from eo-parser).

On top of that, two of them declare a wrong stylesheet id:

- `src/main/resources/org/eolang/parser/add-cuts.xsl:6` -> `id="add-refs"` (should be `add-cuts`)
- `src/main/resources/org/eolang/parser/remove-cuts.xsl:6` -> `id="add-refs"` (should be `remove-cuts`)

These files sit in the same classpath path (`org/eolang/parser/`) as the identically-named XSL files shipped inside `eo-parser`. If the pipeline ever starts loading them by name, the plugin copies will silently shadow the original eo-parser stylesheets, changing normalization behavior without any visible error.

## Expected

Either wire the six stylesheets into the XMIR normalization pipeline (they were clearly meant to normalize generated XMIR: add-refs/add-cuts/remove-cuts/vars-float-down/roll-bases/wrap-method-calls), or remove them. At minimum, fix the duplicated `id="add-refs"`.

## Files
- `src/main/resources/org/eolang/parser/add-cuts.xsl:6`
- `src/main/resources/org/eolang/parser/remove-cuts.xsl:6`
- `src/main/resources/org/eolang/parser/{add-refs,roll-bases,vars-float-down,wrap-method-calls}.xsl`

VasilevNStas:
Fixes #1602

## Problem
Six XSL stylesheets under `src/main/resources/org/eolang/parser/` (`add-cuts.xsl`, `add-refs.xsl`,
`remove-cuts.xsl`, `roll-bases.xsl`, `vars-float-down.xsl`, `wrap-method-calls.xsl`) were **never
loaded by any Java code** — a `grep` over `src/**` finds no reference (the pipeline only uses
`StrictXmir` from eo-parser). Two of them declared a wrong stylesheet id:
- `add-cuts.xsl:6` → `id="add-refs"` (should be `add-cuts`)
- `remove-cuts.xsl:6` → `id="add-refs"` (should be `remove-cuts`)

They sit in the same classpath path as eo-parser's stylesheets, so if the pipeline ever loaded them by
name, the plugin copies would silently shadow the originals

## Solution / What
Removed the six dead stylesheets. Verified they are referenced nowhere in `src/**` (Java or tests), and
that eo-parser 0.63.0 keeps its own pipeline files under `org/eolang/parser/parse/` (different paths —
no shadowing happens today; removing the copies removes the hazard and the duplicated `id="add-refs"`)

## Changes
- Removed `src/main/resources/org/eolang/parser/{add-cuts,add-refs,remove-cuts,roll-bases,vars-float-down,wrap-method-calls}.xsl`

## Tests
- Full `mvn clean install -Pqulice` green (all tests + qulice + jacoco + jtcop, no dependency problems)
- Related: wiring the XMIR normalization pipeline back is tracked separately in #1620


VasilevNStas:
@yegor256 
please review this pull request

yegor256:
CI is green and the reasoning is careful, but I am not deleting six stylesheets on my own judgement. `grep` over `src/**` proves no Java loads them by name today; it does not prove nothing resolves them off the classpath at runtime, and they sit under `org/eolang/parser/`, the same package eo-parser publishes into. If that is wrong the failure shows up in an assemble, not in a unit test.

Assigning to @yegor256, who will know whether those copies were ever load-bearing.

---

PR #1647, opened 2026-09-07
fix(#1632): correct the misleading 'long' profile comment about integration tests

VasilevNStas:
Fixes #1632.

## Problem
The `long` profile comment in `pom.xml` claimed "All integration tests are enabled in the 'long'
profile", while the `pomExcludes` still excluded `phi-unphi/pom.xml` and `jna/pom.xml` (only
`spring-fat` is re-enabled). See also #1605 and #1606.

## Solution / What
Made the comment match reality: all integration tests except `phi-unphi` and `jna` are enabled in the
`long` profile. Re-enabling the two ITs is a separate concern (#1605, #1606), so only the misleading
comment is corrected here.

## Changes
- `pom.xml` — fixed the `long` profile comment.

## Tests
- XML well-formed; no behavior change.

PR #1650, opened 2026-09-07
fix(#1605): re-enable the jna integration test in the long profile

### LINKED ISSUE #1605: The 'jna' integration test is excluded from all profiles and never runs in CI
## What happens

The `jna` integration test (`src/it/jna/`) — the only one exercising native calls (JNA/libc) and the "unpack -> transform -> repack" of a real external library — is excluded from **both** the default and the `long` Maven profiles:

- default profile: `pom.xml:360-372` -> `<exclude>jna/pom.xml</exclude>`
- `long` profile: `pom.xml:565-585` -> `<exclude>jna/pom.xml</exclude>`

A grep over `.github/workflows/*.yml` shows that no workflow ever sets `-Dinvoker.test=jna` (the only `invoker.test` used in CI is `phi-unphi`). So the test never executes anywhere and is effectively dead code.

Additionally, `src/it/jna/pom.xml:26-134` contains a `short` profile (`activeByDefault`) that is never activated because invoker always forces `profiles=full` — a dead duplicate.

## Expected

Either re-enable `jna` (e.g. include it in the `long` profile, or add a dedicated workflow like `phino.yml`), or remove the test and its `short`/`full` profile duplication.

## Files
- `pom.xml:360-372` (default `pomExcludes`)
- `pom.xml:565-585` (`long` `pomExcludes`)
- `src/it/jna/pom.xml:26-134`

VasilevNStas:
Fixes #1605

## Problem
The `jna` integration test — the only one exercising a real external library (JNA/libc) via
unpack → transform → repack — was excluded from **both** the default and the `long` profiles, and no
workflow ever ran it. It was effectively dead code. Additionally, `src/it/jna/pom.xml` had a
`short`/`full` profile duplication, where invoker always forced `full`.

## Solution / What
Re-enabled `jna` in the `long` profile using the **`short`** profile (jeo-only: unpack → disassemble →
assemble → run), which needs only Java 8 — no `eo-maven-plugin`/Docker/Java 17:
1. `src/it/jna/invoker.properties` — `invoker.profiles=full` → `short`; dropped the `invoker.java.version = 17+` requirement.
2. `src/it/jna/pom.xml` — removed the dead `full` profile (it duplicated the phi-unphi pipeline, covered by the `phi-unphi` IT).
3. `pom.xml` (`long` profile) — removed the `jna/pom.xml` exclusion and updated the comment.

## Changes
- `pom.xml` — long profile now runs `jna` (only `phi-unphi` stays excluded).
- `src/it/jna/invoker.properties`, `src/it/jna/pom.xml` — switch to the `short` profile, drop the dead `full` profile.

## Tests
- Ran the IT through the invoker in the long profile:
  `mvn clean install -P"long" -Dinvoker.test=jna` → `jna/pom.xml ... SUCCESS (12.21 s)`, `verify.groovy`
  confirms `Hello from JNA!`.
- Note: this PR and #1647 both touch the `long` profile comment in `pom.xml`; a trivial rebase may be needed for whichever merges second.


VasilevNStas:
@yegor256 
plz review

yegor256:
CI is green, but I'd rather a maintainer took this one. Re-enabling the IT is clearly right; deleting the `full` profile alongside it removes the phi-unphi pipeline from this fixture on the argument that `phi-unphi` covers it, and that is a coverage judgement rather than a defect fix. There is also the noted `long`-profile overlap with #1647. Assigning to @yegor256.

---

zaguzovmaksim0-hue (changes requested):
The re-enable itself looks correct: the `short` profile still unpacks JNA, disassembles the external classes, assembles them back, and `verify.groovy` checks the resulting application actually prints `Hello from JNA!`.

I don't think deleting `full` is justified as duplication in this PR, though. `full` uniquely sends the JNA disassembly through `eo-maven-plugin:xmir-to-phi` and `phi-to-xmir` before reassembly. The separate `phi-unphi` IT shown here uses `phino rewrite --input=xmir --output=xmir`, so it does not cover that same EO-plugin round trip, and it does not exercise it on JNA bytecode. Removing `full` therefore drops distinct integration coverage unrelated to the defect in #1605.

Please keep/re-enable the JNA test without deleting that distinct profile, or split the profile removal into a separate change where the coverage tradeoff can be decided explicitly.

PR #1651, opened 2026-09-07
test(#1618): fix the vacuous empty-container assertion and document the uniform scheme

### LINKED ISSUE #1618: Empty containers in method XMIR are inconsistent: some are emitted, some are omitted
## What happens

For a method with no annotations/try-catch blocks, the generated XMIR is **inconsistent**:

- empty `params`, `annotations` and `trycatchblocks` containers **are** emitted (`DirectivesMethodTest.java:119-133`):
  - `convertsToEmptyAnnotationsIfThereAreNoAnnotations` expects `./o[@name='annotations']`
  - `containsTryCatchBlocksIfThereAreNoExceptions` expects `./o[@name='trycatchblocks']`
- but empty `body`, `exceptions` and `attributes` are **not** emitted (`DirectivesMethodTest.java:64-78` — `createsEmptyXmirIfMethodIsEmpty` expects their absence).

This is an undocumented, asymmetric scheme: a consumer of the XMIR cannot tell whether a missing `body`/`exceptions`/`attributes` means "empty" or "not applicable", while `annotations`/`trycatchblocks` are always present.

## Expected

Pick one rule and apply it uniformly: either emit all containers (even empty) for a method, or omit all of them — and document the choice (see also the related object-form discussion in `DirectivesMethod.java`).

## Files
- `src/test/java/org/eolang/jeo/representation/directives/DirectivesMethodTest.java:64-78, 96-133`
- `src/main/java/org/eolang/jeo/representation/directives/DirectivesMethod.java`

VasilevNStas:
Fixes #1618

## Problem
The issue claimed an asymmetry: some method containers are emitted when empty (`annotations`,
`trycatchblocks`) while others are omitted (`body`, `exceptions`, `attributes`). Investigation shows
the code **already** emits every method container uniformly (all present as empty `seq.of0` for an
empty method). The apparent asymmetry came from the **vacuous** test `createsEmptyXmirIfMethodIsEmpty`
(`DirectivesMethodTest.java:64-78`), which asserted absence of `@as='body'` etc. — but the containers
use `@name`, so the assertion matched nothing and was always true

## Solution / What
- Replaced the vacuous assertion with a meaningful one: an empty method emits **all** containers
  (`annotations`, `body`, `trycatchblocks`, `attributes`) as empty `seq.of0`
- Documented the uniform rule in the `DirectivesMethod` javadoc (a consumer can rely on the presence
  of every container)

## Changes
- `src/test/java/org/eolang/jeo/representation/directives/DirectivesMethodTest.java` — fixed
  `createsEmptyXmirIfMethodIsEmpty` to assert presence of all empty containers (was `@as`-based and vacuous)
- `src/main/java/org/eolang/jeo/representation/directives/DirectivesMethod.java` — javadoc note about
  the uniform emission

## Tests
- `DirectivesMethodTest` passes (6 tests); the new assertion fails if any container is missing

### DIFF
--- a/src/main/java/org/eolang/jeo/representation/directives/DirectivesMethod.java
+++ b/src/main/java/org/eolang/jeo/representation/directives/DirectivesMethod.java
@@ -32,6 +32,9 @@ import org.xembly.Directives;
  * - Try-catch blocks
  * - Default value (for annotation methods)
  * - And other attributes
+ * <p>All method containers (annotations, body, trycatchblocks, attributes, exceptions, params,
+ * maxs) are always emitted, even when empty — an empty container is represented as an empty
+ * sequence (e.g. {@code seq.of0}). A consumer can rely on the presence of every container.</p>
  * @since 0.1
  */
 public final class DirectivesMethod implements Iterable<Directive> {
--- a/src/test/java/org/eolang/jeo/representation/directives/DirectivesMethodTest.java
+++ b/src/test/java/org/eolang/jeo/representation/directives/DirectivesMethodTest.java
@@ -10,7 +10,6 @@ import org.eolang.jeo.representation.NumberedName;
 import org.eolang.jeo.representation.bytecode.BytecodeMethod;
 import org.eolang.jeo.representation.bytecode.JavaCodec;
 import org.hamcrest.MatcherAssert;
-import org.hamcrest.Matchers;
 import org.junit.jupiter.api.Test;
 import org.objectweb.asm.Opcodes;
 import org.xembly.ImpossibleModificationException;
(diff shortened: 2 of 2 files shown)


VasilevNStas:
@yegor256 
plz review this

zaguzovmaksim0-hue (approved):
I checked the claimed uniform-container behavior against the production emitters. `DirectivesMethod.iterator()` always emits method properties, annotations, body, try/catch blocks and attributes. `DirectivesMethodProperties.iterator()` in turn always appends `exceptions`, `max` and `params`, even when their underlying collections/defaults are empty. So the new documentation describes the actual construction path rather than inventing a new contract.

The old `@as` XPath was indeed vacuous for containers named with `@name`; the replacement assertions exercise real `seq.of0` nodes and will fail if those containers disappear. The change is test/documentation-only and CI is green. Approving.

0crat:
@zaguzovmaksim0-hue Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (30) hits-of-code; +4 to give you at least something.  Your running score is [+44](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1652, opened 2026-09-07
test(#1630): clarify class signature vs supername with realistic data and a generic round-trip test

### LINKED ISSUE #1630: XmlClass mixes class signature and supername in its data model
## What happens

`XmlClass` (`src/main/java/org/eolang/jeo/representation/xmir/XmlClass.java`) reads the class-level `signature` and passes it through `BytecodeClassProperties(access, signature, supername, interfaces)`. In tests (`XmlClassPropertiesTest.java:26-45`) a *type* signature `Ljava/util/List<...>;` is used as the class signature.

This mixes two distinct JVM concepts:
- the **class signature** (`ClassSignature`) is the generic signature of the whole class (e.g. `Ljava/lang/Object;Ljava/util/List<Ljava/lang/String;>;`);
- the **supername** is the direct superclass (e.g. `java/lang/Object`).

`XmlClass` and `BytecodeClassProperties` currently thread these through loosely, and a wrong value can silently slip into the generated bytecode (the model does not distinguish them clearly).

## Expected

Clarify the model: keep `signature` (generic) and `supername` as separate, validated fields in `BytecodeClassProperties`/`XmlClassProperties`, and add a round-trip test with a real generic class (`class Foo<T> extends ArrayList<T>`).

## Files
- `src/main/java/org/eolang/jeo/representation/xmir/XmlClass.java`
- `src/main/java/org/eolang/jeo/representation/bytecode/BytecodeClassProperties.java`
- `src/test/java/org/eolang/jeo/representation/xmir/XmlClassPropertiesTest.java:26-45`

VasilevNStas:
Fixes #1630

## Problem
`XmlClassPropertiesTest` used a **type** signature (`Ljava/util/List<Ljava/lang/String;>;`) as the class
signature together with an unrelated supername (`some/custom/Supername`), which obscured the difference
between the two JVM concepts (class generic signature vs direct superclass). The data model itself keeps
them as separate fields (`BytecodeClassProperties.signature()` / `supername()`, threaded separately into
`ClassVisitor.visit`), so the fix is to document and verify that separation properly

## Solution / What
- Fixed the existing test to use a realistic, consistent class: `class Foo extends ArrayList<String>
  implements List<String>` → signature `Ljava/util/ArrayList<Ljava/lang/String;>;`, supername
  `java/util/ArrayList`, interfaces `{java/util/List}`
- Added a full round-trip test with a generic class including a type parameter:
  `class Foo<T extends Number> extends ArrayList<T>` → signature `<T::Ljava/lang/Number;>Ljava/util/ArrayList<TT;>;`
  The test builds a `BytecodeObject`, serializes it to XMIR, parses it back via `XmlObject`, and asserts
  the signature and supername survive separately

## Changes
- `src/test/java/org/eolang/jeo/representation/xmir/XmlClassPropertiesTest.java` — realistic data in
  `createsXmirWithCorrectProperties` + new `keepsClassSignatureSeparateFromSupername`

## Tests
- `XmlClassPropertiesTest` passes (2 tests), including the generic round-trip

### DIFF
--- a/src/test/java/org/eolang/jeo/representation/xmir/XmlClassPropertiesTest.java
+++ b/src/test/java/org/eolang/jeo/representation/xmir/XmlClassPropertiesTest.java
@@ -4,7 +4,10 @@
  */
 package org.eolang.jeo.representation.xmir;
 
+import java.util.ArrayList;
+import org.eolang.jeo.representation.bytecode.BytecodeClass;
 import org.eolang.jeo.representation.bytecode.BytecodeClassProperties;
+import org.eolang.jeo.representation.bytecode.BytecodeObject;
 import org.eolang.jeo.representation.directives.DirectivesClassProperties;
 import org.hamcrest.MatcherAssert;
 import org.hamcrest.Matchers;
@@ -20,9 +23,9 @@ final class XmlClassPropertiesTest {
     @Test
     void createsXmirWithCorrectProperties() {
         final int access = Opcodes.ACC_PUBLIC | Opcodes.ACC_ABSTRACT | Opcodes.ACC_SUPER;
-        final String signature = "Ljava/util/List<Ljava/lang/String;>;";
-        final String supername = "some/custom/Supername";
-        final String[] interfaces = {"java/util/List", "java/util/Collection"};
+        final String signature = "Ljava/util/ArrayList<Ljava/lang/String;>;";
+        final String supername = "java/util/ArrayList";
+        final String[] interfaces = {"java/util/List"};
         MatcherAssert.assertThat(
             "We expect that the properties will be created correctly and contain the correct values",
             new XmlClass(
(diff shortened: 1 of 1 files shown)


VasilevNStas:
@yegor256 
plz review

zaguzovmaksim0-hue (changes requested):
The round-trip mechanism is useful, but the new test's claimed "realistic" generic class signature does not represent the Java class described in the PR.

I compiled exactly this on JDK 21:

```java
import java.util.ArrayList;
class Foo<T extends Number> extends ArrayList<T> {}
```

and `javap -v` reports:

```
Signature: <T:Ljava/lang/Number;>Ljava/util/ArrayList<TT;>;
```

The PR uses `<T::Ljava/lang/Number;>...`. In the JVM signature grammar the extra colon means the class bound is absent and `Number` occupies an interface-bound slot; it is not the signature javac emits for `T extends Number`. Because this test only serializes and parses the string, it can round-trip an unrealistic/semantically different signature and still pass.

Please change the fixture to `<T:Ljava/lang/Number;>Ljava/util/ArrayList<TT;>;` (or change the described source type to match the double-colon form) so the test actually establishes the class-signature/supername separation using valid real-world data.

PR #1653, opened 2026-09-07
test(#1603): cross-check OpcodeDictionary against ASM opcode constants

### LINKED ISSUE #1603: Add a verification test that every JVM opcode is mapped in OpcodeDictionary
## What happens

`OpcodeDictionary` (`src/main/java/org/eolang/jeo/representation/OpcodeDictionary.java`) maps opcode numbers to names and back, but it is missing several legal JVM opcodes:

- `wide` (196), `goto_w` (200) — the table stops at `ifnonnull` (199);
- `ldc_w` (19), `ldc2_w` (20);
- the indexed load/store forms `iload_n`/`lload_n`/`fload_n`/`dload_n`/`aload_n` (26-45) and `istore_n`/... (59-78).

If an XMIR contains any of these names, `assemble` fails with `UnrecognizedOpcode`, while `disassemble` normalizes them away. `OpcodeDictionaryTest` has only 5 tests and never cross-checks the dictionary against ASM's `org.objectweb.asm.Opcodes`, which is why these gaps are unnoticed.

## Expected

Add a parameterized verification test that iterates over the JVM opcode constants in `org.objectweb.asm.Opcodes` and asserts that `code(name(op)) == op` for every supported opcode (and, for the missing ones, decide whether to add them to the dictionary or explicitly document them as unsupported). The indexed `iload_n`/`istore_n` forms should be added since `BytecodeInstruction` already emits them.

## Files
- `src/main/java/org/eolang/jeo/representation/OpcodeDictionary.java:128-288`
- `src/test/java/org/eolang/jeo/representation/OpcodeDictionaryTest.java:18-62`

VasilevNStas:
Fixes #1603.

## Problem
`OpcodeDictionary` is missing some legal JVM opcode names (`ldc_w`, `ldc2_w`, `iload_n`..`aload_n`,
`istore_n`..`astore_n`, `wide`, `goto_w`, `jsr_w`), and `OpcodeDictionaryTest` never cross-checked the
dictionary against ASM's `org.objectweb.asm.Opcodes`.

## Investigation
ASM **normalizes** all the "missing" opcodes and never exposes them as constants:
- `ldc_w`/`ldc2_w` → ASM emits `LDC` and picks the wide form automatically;
- `iload_0`..`aload_3` / `istore_0`..`astore_3` → ASM emits `ILOAD`/`ISTORE` + index;
- `wide` → ASM widens the next instruction internally;
- `goto_w`/`jsr_w` → ASM emits `GOTO`/`JSR` (`jsr` is banned in class files ≥ 52).

No code path emits them, and no `.xmir` resource contains them. Adding them to the dictionary alone
would be misleading (`code("iload_0")` would resolve but `assemble` would still fail at
`Instruction.find(26)` → `UnrecognizedOpcode`). So they are explicitly documented as unsupported.

## Solution / What
- Added a parameterized test iterating the **genuine** opcode constants of `org.objectweb.asm.Opcodes`
  (filtered from `ACC_*`/`V1_*`/`TOP`/... by name match) and asserting `code(name(op)) == op` for each —
  a drift guard that fails if the dictionary loses or misnames an ASM-visible opcode.
- Added a test asserting the ASM-normalized pseudo-opcodes are rejected, documenting the decision.

## Changes
- `src/test/java/org/eolang/jeo/representation/OpcodeDictionaryTest.java` — 2 new parameterized tests
  + the ASM-constant enumeration source.

## Tests
- `OpcodeDictionaryTest` passes (172 tests; 157 genuine ASM opcode constants verified).
- Full `mvn clean install -Pqulice` green.

### DIFF
--- a/src/test/java/org/eolang/jeo/representation/OpcodeDictionaryTest.java
+++ b/src/test/java/org/eolang/jeo/representation/OpcodeDictionaryTest.java
@@ -4,10 +4,19 @@
  */
 package org.eolang.jeo.representation;
 
+import java.lang.reflect.Field;
+import java.util.Arrays;
+import java.util.Locale;
+import java.util.stream.Stream;
 import org.hamcrest.MatcherAssert;
 import org.hamcrest.Matchers;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.junit.jupiter.params.provider.ValueSource;
+import org.objectweb.asm.Opcodes;
 
 /**
  * Test case for {@link OpcodeDictionary}.
@@ -19,7 +28,7 @@ final class OpcodeDictionaryTest {
     void retrievesNameForValidOpcode() {
         MatcherAssert.assertThat(
             "We expect that the opcode name for NOP is 'nop'",
-            new OpcodeDictionary().name(org.objectweb.asm.Opcodes.NOP),
+            new OpcodeDictionary().name(Opcodes.NOP),
             Matchers.is(Matchers.equalTo("nop"))
         );
     }
@@ -38,7 +47,7 @@ final class OpcodeDictionaryTest {
         MatcherAssert.assertThat(
             "We expect that the opcode for 'nop' is correct",
             new OpcodeDictionary().code("nop"),
-            Matchers.is(Matchers.equalTo(org.objectweb.asm.Opcodes.NOP))
+            Matchers.is(Matchers.equalTo(Opcodes.NOP))
         );
     }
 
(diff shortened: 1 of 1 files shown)


VasilevNStas:
@yegor256 
plz review

src/test/java/org/eolang/jeo/representation/OpcodeDictionaryTest.java
```diff
@@ -56,7 +65,91 @@ void handlesCaseInsensitiveOpcodeNames() {
...
+     * Check that the field is a genuine opcode constant.
+     * @param field Field to check.
+     * @return True if the field is an opcode constant.
+     */
+    private static boolean isAsmOpcode(final Field field) {
+        final int code = OpcodeDictionaryTest.value(field);
+        boolean result = false;
+        if (code >= Opcodes.NOP && code <= Opcodes.IFNONNULL) {
+            try {
+                result = new OpcodeDictionary().name(code)
```
zaguzovmaksim0-hue, line 131: This filter makes the new drift guard circular. If OpcodeDictionary is missing a genuine ASM opcode, name(code) throws here, the catch turns result=false, and that opcode is excluded from asmOpcodes() — so mapsEveryAsmOpcodeBackToItself never sees precisely the missing entry it is supposed to detect. The candidate opcode set needs to be identified independently of OpcodeDictionary (for example from a fixed ASM opcode-name/range source), then the dictionary should be asserted against it.

zaguzovmaksim0-hue (changes requested):
The pseudo-opcode decision may be reasonable, but the advertised ASM cross-check currently cannot detect dictionary omissions because the dictionary itself decides which ASM fields enter the test. Please make the candidate set independent of OpcodeDictionary so deleting/missing a supported opcode actually fails the test.

PR #1657, opened 2026-09-07
test(#1614): add an integration test that fails on invalid XMIR with xmirVerification enabled

### LINKED ISSUE #1614: Add an integration test that fails on invalid XMIR with xmirVerification enabled
## What happens

`xmirVerification` (the plugin's own XMIR validation via `objectionary/lints`) is only exercised in the `daily` workflow (`-Pverify`, where `verify.xmir.in.integration.tests=true`), and no integration test covers its **negative** path:

- In the default `mvn clean install`, `verify.xmir.in.integration.tests` is `false` (`pom.xml:72`), so the verification code path is not covered at all by the IT suite.
- There is no IT that feeds an **invalid/corrupted XMIR** into `assemble` and asserts the build fails with a clear error.

The `annotations` IT even has the verification-related assertions commented out with a `// todo: enable` (`src/it/annotations/verify.groovy:10-11`).

## Expected

Add an integration test that:
- enables `xmirVerification` (e.g. `-Djeo.assemble.xmir.verification=true` or via the `verify` profile),
- passes a deliberately invalid XMIR to `assemble`,
- asserts the build fails with an informative `IllegalArgumentException`/`MojoExecutionException`.

This locks in the documented option and closes the negative scenario.

## Files
- `src/it/annotations/verify.groovy:10-11`
- `pom.xml:72` (`verify.xmir.in.integration.tests` default)
- `src/it/annotations/pom.xml:31` (xmirVerification wiring)

VasilevNStas:
Fixes #1614 

## Problem
The `xmirVerification` option (XMIR validation via `objectionary/lints`) was only exercised in the
`daily` workflow (`-Pverify`), and no integration test covered its **negative** path: feeding an
invalid/corrupted XMIR into `assemble` and asserting the build fails

## Solution / What
Added a dedicated integration test `src/it/xmir-verification/`:
- `pom.xml` — runs the `assemble` goal with `<xmirVerification>true</xmirVerification>` and
  `sourcesDir` pointing at `src/xmir`
- `src/xmir/Invalid.xmir` — a deliberately invalid XMIR (`<program></program>`, invalid per the XMIR schema).
- `invoker.properties` — `invoker.buildResult=failure` (the build is expected to fail)
- `verify.groovy` — asserts the log contains `BUILD FAILURE`, the "Verifying all the XMIR files
  before assembling" message, and the schema error `Cannot find the declaration of element 'program'`

## Changes
- New IT under `src/it/xmir-verification/` (covered by the default `*/pom.xml` invoker includes)

## Tests
- `mvn clean install -Dinvoker.test=xmir-verification` passes (build fails as expected, verify script green)
- Full `mvn clean install -Pqulice` green (all tests + all ITs + qulice + jacoco)


VasilevNStas:
@yegor256 
plz review

src/it/xmir-verification/verify.groovy
```diff
@@ -0,0 +1,17 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+ * SPDX-License-Identifier: MIT
+ */
+String log = new File(basedir, 'build.log').text;
+assert log.contains("BUILD FAILURE"): assertionMessage("BUILD FAILURE not found")
+assert log.contains("Verifying all the XMIR files before assembling"): assertionMessage("XMIR verification was not started")
+assert log.contains("Cannot find the declaration of element 'program'"): assertionMessage("The invalid XMIR was not reported")
```
zaguzovmaksim0-hue, line 8: This pins a parser/schema-vendor diagnostic that is not jeo's contract. PR #1645 in this same repository specifically removes an XSD assertion like this because Xerces/jcabi wording can change while validation remains correct. The negative IT can robustly assert BUILD FAILURE plus the plugin's own 'Verifying all the XMIR files before assembling' marker (and, if needed, a stable jeo-owned exception prefix) without depending on 'Cannot find the declaration...' verbatim.

zaguzovmaksim0-hue (changes requested):
The new negative integration coverage is useful, but it reintroduces a brittle dependency on exact XSD-validator wording. Please assert a stable plugin-owned failure signal instead of the Xerces/jcabi message.

PR #1658, opened 2026-09-07
test(#1617): add integration tests for mojo parameters and the negative threads value

### LINKED ISSUE #1617: Most mojo parameters have no dedicated integration test (positive or negative)
## What happens

Only a few mojo parameters are covered by integration tests. Most have no positive or negative IT at all:

- `threads < 0` -> `MojoExecutionException` (we have a unit test in `ParallelTranslatorTest`, but no IT proves the user-visible behavior);
- `disabled=true` -> goal does nothing;
- `skipVerification` (assemble) -> skips bytecode verification;
- `debug` / `mode=debug`;
- `omitListings`, `omitComments`, `prettyXmir`, `modifiers`.

Current ITs that touch some of these (`src/it/bytecode-to-eo`, `spring-fat`, `scala`, `scala2`) use them incidentally, not as a dedicated scenario. A regression in e.g. `disabled` or `skipVerification` would go unnoticed.

## Expected

Add integration tests (one IT or a set) covering the negative and positive behavior of the parameters: e.g. `threads=-1` fails with a clear `MojoExecutionException`, `disabled=true` produces no output, `skipVerification=true` skips the JVM check, `omitComments`/`omitListings`/`prettyXmir` change the generated XMIR.

## Files
- `src/it/configurable-threads/pom.xml:37-49` (threads wiring)
- `src/main/java/org/eolang/jeo/DisassembleMojo.java` (parameters)
- `src/main/java/org/eolang/jeo/AssembleMojo.java` (parameters)

VasilevNStas:
Fixes #1617

## Problem
Most mojo parameters had no dedicated integration test (positive or negative). A regression in e.g.
`disabled`, `skipVerification`, `mode`, `omitComments`/`omitListings`, `modifiers` or the negative
`threads` value would go unnoticed.

## Solution / What
Added two integration tests:

**`src/it/mojo-parameters/`** (positive) — one project with four plugin executions, each verified in `verify.groovy`:
- `disassemble-debug`: `mode=debug`, `omitComments=false`, `omitListings=false`, `prettyXmir=false`
  → XMIR contains `<listing>` and `<!--` comments;
- `disassemble-modifiers`: `modifiers=true` → XMIR contains method `modifiers`;
- `disassemble-disabled`: `disabled=true` → no `jeo-xmir-disabled` output and the "disabled" log line;
- `assemble-skip-verification`: `skipVerification=true` → "Bytecode verification is disabled" log line
  and the assembled `.class` exists.

**`src/it/negative-threads/`** (negative) — `threads=-1` with `invoker.buildResult=failure`; `verify.groovy`
asserts `BUILD FAILURE` and the descriptive message `The number of threads must be 0 or positive, but got: -1`.

## Changes
- Two new ITs under `src/it/` (covered by the default `*/pom.xml` invoker includes).

## Tests
- Both ITs pass via `mvn clean install -Dinvoker.test=...`.
- Full `mvn clean install -Pqulice` green (all tests + all ITs + qulice + jacoco).


VasilevNStas:
@yegor256 
plz review this PR

zaguzovmaksim0-hue (changes requested):
The new IT covers several parameters well, but two parameters named by #1617 are configured without testing their behavior: `prettyXmir` and `mode`.

In `DisassembleMojo`, `prettyXmir` is passed independently as `Format.PRETTY`; `BytecodeRepresentation.toEO()` then chooses `PrettyXml` only when that flag is true. The new execution sets `<prettyXmir>false</prettyXmir>`, but `verify.groovy` never checks that the output is non-pretty. If the mojo stopped wiring this parameter and silently used the default `true`, this IT would still pass.

The same problem exists for `<mode>debug</mode>`. The assertions for `<listing>` and comments exercise `omitListings=false` and `omitComments=false`, which are separate `Format` properties. They do not distinguish DEBUG (`ClassReader` flags = 0) from SHORT (`SKIP_DEBUG`). If the `mode` parameter were ignored and defaulted to `short`, the current assertions can still pass.

Please add assertions that independently fail when `prettyXmir=false` is ignored and when `mode=debug` is ignored (for example, a short-vs-debug comparison/debug-only metadata for `mode`, and compact-vs-pretty serialization for `prettyXmir`). That will make the IT actually guard the two parameter wirings the issue asks us to cover.

PR #1659, opened 2026-09-08
feat(#1604): make the generated XMIR 'time' attribute reproducible

### LINKED ISSUE #1604: Make generated XMIR reproducible by making the 'time' attribute configurable
## What happens

`DirectivesObject` writes a wall-clock timestamp into every generated XMIR:

```java
// src/main/java/org/eolang/jeo/representation/directives/DirectivesObject.java:89-97
final String now = ZonedDateTime.now(ZoneOffset.UTC)...
   .attr("time", now)
```

Because of this, the output of `disassemble` is **not byte-reproducible between runs**: running the same build twice produces XMIR files that differ only in `<program time="...">`. This blocks reproducible/hermetic builds (e.g. comparing artifacts, caching, verification of generated sources in CI).

## Expected

Make the timestamp optional/configurable. Options:
- a `Format` flag (e.g. `Format.REPRODUCIBLE`) that omits or fixes `time`;
- honor an environment/property (e.g. `SOURCE_DATE_EPOCH`, the standard for reproducible builds) when present.

The default behavior can stay unchanged.

## Files
- `src/main/java/org/eolang/jeo/representation/directives/DirectivesObject.java:89-97`
- `src/main/java/org/eolang/jeo/representation/directives/Format.java`

VasilevNStas:
Fixes #1604

## Problem
`DirectivesObject` wrote a wall-clock timestamp into every generated XMIR (`<program time="...">`),
so the `disassemble` output was not byte-reproducible between runs

## Solution / What
Made the timestamp configurable, with three sources in priority order:
1. **`Format.TIME`** property — an explicit ISO-8601 timestamp (full control, testable);
2. **`SOURCE_DATE_EPOCH`** environment variable — the standard for reproducible builds (Unix epoch
   in seconds), honored automatically when present;
3. current wall-clock time — unchanged default behavior

Implementation:
- `Format` — new `TIME` constant and `time()` accessor
- `DirectivesObject` — new package-private `time(Format, String epoch)` helper (used by `iterator()`)
  plus `current()` for the default

## Changes
- `src/main/java/org/eolang/jeo/representation/directives/Format.java`
- `src/main/java/org/eolang/jeo/representation/directives/DirectivesObject.java`

## Tests
- `DirectivesObjectTest` — `usesConfiguredTime` (XMIR has `time="2024-01-01T00:00:00Z"`),
  `usesSourceDateEpoch` (`time(new Format(), "1704067200")` → `2024-01-01T00:00:00Z`),
  `usesCurrentTimeByDefault`

### DIFF
--- a/src/main/java/org/eolang/jeo/representation/directives/DirectivesObject.java
+++ b/src/main/java/org/eolang/jeo/representation/directives/DirectivesObject.java
@@ -5,10 +5,12 @@
 package org.eolang.jeo.representation.directives;
 
 import com.jcabi.manifests.Manifests;
+import java.time.Instant;
 import java.time.ZoneOffset;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeFormatter;
 import java.util.Iterator;
+import java.util.Optional;
 import org.xembly.Directive;
 import org.xembly.Directives;
 
@@ -86,8 +88,9 @@ public final class DirectivesObject implements Iterable<Directive> {
 
     @Override
     public Iterator<Directive> iterator() {
-        final String now = ZonedDateTime.now(ZoneOffset.UTC)
-            .format(DateTimeFormatter.ISO_INSTANT);
+        final String now = DirectivesObject.time(
+            this.format, System.getenv("SOURCE_DATE_EPOCH")
+        );
         final Directives directives = new Directives()
             .add("object")
             .attr("xmlns:xsi", "http://www.w3.org/2001/XMLSchema-instance")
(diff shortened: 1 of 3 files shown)


VasilevNStas:
@yegor256 
plz review

zaguzovmaksim0-hue (approved):
The precedence and reproducibility behavior match #1604. A configured `Format.TIME` wins, otherwise a non-empty `SOURCE_DATE_EPOCH` is converted from epoch seconds to an ISO instant, and only the absence of both falls back to the existing UTC wall-clock timestamp.

For Maven users the standard environment variable path requires no new Mojo parameter, so reproducible builds are actually reachable through the normal plugin execution; the `Format.TIME` option additionally makes the mechanism deterministic for direct/programmatic use and tests. The known epoch fixture maps correctly to `2024-01-01T00:00:00Z`, and the default path preserves current behavior. Full CI is green. Approving.

0crat:
@zaguzovmaksim0-hue Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+36](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1660, opened 2026-09-08
fix(#1610): align the README Maven version requirement with the actual minimum

### LINKED ISSUE #1610: Plugin builds against Maven 4.0.0-alpha-7 API but README promises Maven 3.1+
## What happens

The plugin is compiled against `org.apache.maven:maven-plugin-api:4.0.0-alpha-7` (`pom.xml:147`, scope `provided`), i.e. the Maven 4 API line. But the README documents the requirement as:

- `README.md:20` — "Maven 3.1+"
- `README.md:600` — "Maven 3.3+" (a second, different statement)

Running the plugin on an actual Maven 3.x may fail at runtime with `NoSuchMethodError`/`NoClassDefFoundError` if any API method/class from the 4.0 line is used (the plugin is also built with the jcabi parent that sets Maven prerequisites). The two README claims contradict each other, and neither matches the API level the code is compiled against.

## Expected

Decide the real minimum Maven version and make it consistent: either compile against `maven-plugin-api` 3.x (if Maven 3.1/3.3 support is intended) or raise the documented requirement to Maven 4 and align both README statements.

## Files
- `pom.xml:147` (`maven-plugin-api` 4.0.0-alpha-7)
- `README.md:20` vs `README.md:600`

VasilevNStas:
Fixes #1610

## Problem
The README contradicted itself about the minimum Maven version:
- `README.md:20` — "Maven 3.1+";
- `README.md:600` (the `-Pverify` section) — "Maven 3.3+"

Meanwhile the plugin is compiled against `maven-plugin-api:4.0.0-alpha-7` (an intentional choice for
Java 8 compatibility, see the `pom.xml` comment)

## Investigation
The plugin only uses the **base** Maven API (`AbstractMojo`, `MojoExecutionException`, `@Mojo`,
`@Parameter`, `LifecyclePhase`, and `MavenProject#get*ClasspathElements()` from `maven-core`), all of
which exist and are stable in the Maven 3.x line. Nothing specific to the Maven 4 API is used, so the
plugin runs on Maven 3.x (the whole build and IT suite are executed on Maven 3.9.16 locally and in CI)

## Solution / What
Unified the documented minimum to **Maven 3.1+** in both places (matching the primary statement)
`maven-plugin-api:4.0.0-alpha-7` is kept unchanged — switching it would risk the Java 8
compatibility the `pom.xml` comment explicitly warns about

## Changes
- `README.md` — `README:600` now says "Maven 3.1+" instead of "Maven 3.3+"

## Tests
- Documentation-only change; no tests required

### DIFF
--- a/README.md
+++ b/README.md
@@ -597,4 +597,4 @@ It might take significantly more time to build,
 but it will ensure that all transformations are correct and aligned with
 the XMIR specification.
 
-You will need [Maven 3.3+](https://maven.apache.org) and Java 11+ installed.
+You will need [Maven 3.1+](https://maven.apache.org) and Java 11+ installed.


VasilevNStas:
@yegor256 
plz review this

zaguzovmaksim0-hue (changes requested):
This change makes the README internally consistent, but with a minimum Maven version that the built plugin itself rejects.

I checked the published `org.eolang:jeo-maven-plugin:0.15.3` artifact from Maven Central, not just the Java API calls. Its generated `META-INF/maven/plugin.xml` contains:

```
<requiredJavaVersion>11</requiredJavaVersion>
<requiredMavenVersion>4.0.0-alpha-7</requiredMavenVersion>
```

That value comes from the plugin metadata generated from the build, and this PR does not change the `maven-plugin-api:4.0.0-alpha-7` dependency that drives it. Therefore Maven 3.1 is not the supported runtime minimum of the artifact even if the Mojo source happens to use only old-looking API methods. The current CI on Maven 3.9.x also does not establish compatibility with 3.1.

Please either document the generated requirement (`4.0.0-alpha-7+`, and make both README locations agree) or change the build/API dependency so the generated plugin descriptor truly allows Maven 3.x, then test that runtime. Simply changing `3.3+` to `3.1+` leaves #1610's core mismatch unresolved.

Issue #1633, opened 2026-09-06
maven.test.yml runs the full build with integration tests on all 3 OS x 2 JDK; consider splitting


VasilevNStas:
## What happens

`.github/workflows/maven.test.yml:8-20` uses a matrix of 3 OS (`ubuntu-24.04`, `macos-15`, `windows-2022`) × 2 JDK (`11`, `23`) and runs `mvn clean install` — **including the maven-invoker integration tests — on all six combinations**:

```yaml
strategy:
  matrix:
    os: [ ubuntu-24.04, macos-15, windows-2022 ]
    java: [ 11, 23 ]
```

The IT suite (20+ projects) is the slow part and mostly exercises JDK-independent bytecode logic; running it on every OS×JDK combination burns CI minutes, especially on Windows/macOS runners.

## Expected

Consider splitting: run unit tests (`mvn test`) on the full matrix (OS×JDK), but run the invoker ITs only on a single representative combination (e.g. ubuntu + one JDK), or gate ITs behind a separate workflow like `maven.long.yml`.

## Files
- `.github/workflows/maven.test.yml:8-20`

0crat:
@vasilevnstas Thanks for suggesting an enhancement! You've earned +12 points. If you find anything else in the repository that [doesn't look](https://www.yegor256.com/2018/02/06/where-to-find-more-bugs.html) as good as you might expect, [do not hesitate](https://www.yegor256.com/2014/04/13/bugs-are-welcome.html) to [report](https://www.yegor256.com/2018/04/24/right-way-to-report-bugs.html) bugs too. Your running score is [+1458](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1665, opened 2026-09-08
fix(#1633): split CI into unit tests on the full matrix and integration tests on one combination

### LINKED ISSUE #1613: up.yml uses sign-commits with the default GITHUB_TOKEN and a fragile version grep

VasilevNStas:
Fixes #1613 

## Problem
`maven.test.yml` ran `mvn clean install` — including the full invoker IT suite (20+ projects) — on all
six OS×JDK combinations (`ubuntu-24.04`/`macos-15`/`windows-2022` × `11`/`23`). The IT suite is the slow
part and mostly exercises JDK-independent bytecode logic, so this burned CI minutes on every runner

## Solution / What
Split the build:
- **Unit tests** (`mvn clean test`) run on the full OS×JDK matrix (all six combinations)
- **Integration tests** (`mvn clean install -DskipTests`, which still executes the invoker ITs) run only
  on a single representative combination (`ubuntu-24.04` + JDK 23), added via a matrix `include`

The ITs remain fully covered on the representative runner (and additionally by `maven.long.yml` on
ubuntu)

## Changes
- `.github/workflows/maven.test.yml` — two steps with the `it` matrix flag

## Tests
- `yamllint .github/workflows/maven.test.yml` passes; `actionlint` is run by the CI (`actionlint.yml`)

### DIFF
--- a/.github/workflows/maven.test.yml
+++ b/.github/workflows/maven.test.yml
@@ -16,6 +16,10 @@ jobs:
       matrix:
         os: [ ubuntu-24.04, macos-15, windows-2022 ]
         java: [ 11, 23 ]
+        include:
+          - os: ubuntu-24.04
+            java: 23
+            it: true
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v6
@@ -29,4 +33,8 @@ jobs:
           key: ${{ runner.os }}-jdk-${{ matrix.java }}-maven-${{ hashFiles('**/pom.xml') }}
           restore-keys: |
             ${{ runner.os }}-jdk-${{ matrix.java }}-maven-
-      - run: mvn clean install -ntp --errors --batch-mode
+      - name: Run unit tests
+        run: mvn clean test -ntp --errors --batch-mode
+      - name: Run integration tests
+        if: matrix.it
+        run: mvn clean install -ntp --errors --batch-mode -DskipTests


VasilevNStas:
@yegor256 
plz review

zaguzovmaksim0-hue (approved):
I verified the matrix behavior against the actual PR workflow run, not only the YAML. The existing ubuntu-24.04/JDK23 matrix job received `it: true` (no duplicate seventh matrix job), and after the unit-test step its install step ran `maven-invoker-plugin:3.9.1`. The log shows the regular IT projects (`spring`, `records`, `variable-names`, `generics`, `annotations`, `inner-classes`, `exceptions`, `takes`, `custom-transformations`, `bytecode-to-eo`, `excludes-includes`, `configurable-threads`, `scala2`, `modules`, `scala`, and the without-pom cases) all executing and ending SUCCESS.

The other OS/JDK combinations still run unit tests, while the existing mvn-long workflow remains as additional ubuntu integration coverage. The split therefore removes redundant cross-platform IT runs without silently disabling the suite. Approving.

0crat:
@zaguzovmaksim0-hue Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (10) hits-of-code; +4 to give you at least something.  Your running score is [+40](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1676, opened 2026-09-10
#1675: refuse a hex value that is no sequence of bytes

### LINKED ISSUE #1675: `XmlValue` throws a bare `ArrayIndexOutOfBoundsException` on an odd-length hex value
**What happens**

`XmlValue.bytes()` turns the hex text of an XMIR value into a byte array with
no check that the text is actually a well-formed hex pair sequence:

```java
final char[] chars = hex.toCharArray();
final int length = chars.length;
res = new byte[length / 2];
for (int index = 0; index < length; index += 2) {
    res[index / 2] = (byte) Integer.parseInt(
        String.copyValueOf(new char[]{chars[index], chars[index + 1]}), XmlValue.RADIX
    );
}
```

An odd number of hex digits runs off the end of both arrays, and a non-hex
character reaches `Integer.parseInt`. Through the real class:

```
valid  41-42 -> ok: AB
odd    41-4  -> THREW java.lang.ArrayIndexOutOfBoundsException: Index 3 out of bounds for length 3
nonhex ZZ    -> THREW java.lang.NumberFormatException: For input string: "ZZ" under radix 16
```

**Why it is wrong**

Neither message names the file, the class, the method or the instruction the
value belongs to, so there is nothing to go on. This is the bug class this
repository has fixed several times already (#939, #964, #967, #1538): a
low-level accessor throwing a bare technical exception on unusual XMIR where a
descriptive `IllegalStateException` is the established idiom. `XmlValue` itself
follows that idiom two methods away:

```java
() -> new IllegalStateException(
    String.format("Can't find a child in '%s' to convert to hex", this.node)
)
```

**Impact**

XMIR is not only produced by this plugin, it is meant to be rewritten between
`disassemble` and `assemble` - that is what `hone-maven-plugin` does with
phino. A rewrite that emits a malformed byte string, or a hand-edited file,
fails the build with an index out of bounds and no indication of where in the
program it came from.

**Evidence**

The three lines above were produced by calling `XmlValue.object()` against
`target/classes` on `<o base='org.eolang.jeo.string'><o>41-4</o></o>` and
`<o base='org.eolang.jeo.string'><o>ZZ</o></o>`.

**What should happen**

A value whose hex text is not a sequence of byte pairs should be refused with a
message that names the offending text and the node it came from, the way the
missing-child case already is.

morphqdd:
`XmlValue.bytes()` read the hex text with no check on it, so an odd number of
digits ran off the end of the array and a non-hex pair reached `parseInt`.
Neither `ArrayIndexOutOfBoundsException` nor `NumberFormatException` names the
value or the node it came from.

Both cases are now refused with the descriptive `IllegalStateException` the
class already uses two methods away.

Closes #1675


zaguzovmaksim0-hue (approved):
I checked this at the normalized-hex boundary used by `XmlValue.bytes()`: `hex()` removes the EO `-` delimiters first, so the validation here correctly operates on the actual digit sequence that is about to be decoded.

The odd-length guard prevents both the `chars[index + 1]` overrun and the undersized destination-array write before the loop starts. The per-pair `NumberFormatException` wrapper then handles malformed hex without masking the original cause, and both messages include the offending value/node context that the previous raw exceptions lacked. Valid even-length pairs keep the same `Integer.parseInt(..., 16)` conversion.

The two regression tests cover the distinct failure classes (`41-4` -> odd normalized length, `ZZ` -> non-hex pair). No correctness issue found in the change. Approving while the remaining CI matrix completes.

yegor256:
Merged, thanks!

---

0crat:
@zaguzovmaksim0-hue Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+120](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +16 points. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5885](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1677, opened 2026-09-10
#1674: split the encoded signature at its last dash

### LINKED ISSUE #1674: `NamedDescriptor` splits on the first dash, silently corrupting a method name that contains one
**What happens**

`NamedDescriptor` joins a method name and its descriptor with a dash:

```java
public String encoded() {
    return String.format(
        "%s-%s",
        this.original,
        new DecodedString(this.descr).encode()
    );
}
```

and splits them back on the *first* dash:

```java
private static String prefix(final String encoded) {
    ...
    return encoded.substring(0, encoded.indexOf('-'));
}

private static String suffix(final String encoded) {
    return new EncodedString(encoded.substring(encoded.indexOf('-') + 1)).decode();
}
```

The descriptor is URL-encoded, so it can never contain a raw dash and the last
dash would always be the right separator. The name is not encoded, and a dash
is legal in a JVM method name (JVMS 4.2.2 forbids only `.`, `;`, `[`, `/`, and
`<`/`>` for methods), which is what Kotlin emits for value classes:
`box-impl`, `unbox-impl`, `constructor-impl`.

For such a name the split lands inside the name and nothing complains:

```
name=plain     desc=()V   encoded=plain-%28%29V      back.name=plain  back.desc=()V       ok=true
name=box-impl  desc=()V   encoded=box-impl-%28%29V   back.name=box    back.desc=impl-()V  ok=false
```

**Impact**

The name and the descriptor both come back wrong, with no exception. An
annotation whose value refers to such a method is silently rewritten to point
at a method named `box` with the descriptor `impl-()V`, which is not a valid
descriptor and does not name anything that exists.

This is the quiet half of the same problem as #1673: there the mismatch throws,
here it corrupts.

**Evidence**

Output above was produced against `target/classes` by encoding and decoding
through the class itself.

**What should happen**

The split should use the last dash, which the encoding guarantees is the
separator, so that a name containing a dash survives the round trip.

morphqdd:
The name and the descriptor were joined with a dash and split back on the first
one. A dash is legal in a JVM method name and Kotlin emits `box-impl`,
`unbox-impl` and `constructor-impl` for value classes, so the split landed
inside the name and both halves came back wrong with no exception. The
`annotation-N` names this plugin makes itself had the same problem.

The descriptor's own dashes are escaped now, so the last dash is the separator,
and the split uses it.

Closes #1674

### DIFF
--- a/src/main/java/org/eolang/jeo/representation/NamedDescriptor.java
+++ b/src/main/java/org/eolang/jeo/representation/NamedDescriptor.java
@@ -45,13 +45,18 @@ public final class NamedDescriptor {
 
     /**
      * Encoded method name with descriptor.
+     *
+     * <p>A dash of the descriptor is escaped, so the last dash of the result
+     * is the separator. A method name may hold a dash of its own - Kotlin
+     * emits "box-impl" for a value class - and splitting on the first one
+     * landed inside the name.</p>
      * @return Encoded method name with descriptor.
      */
     public String encoded() {
         return String.format(
             "%s-%s",
             this.original,
-            new DecodedString(this.descr).encode()
+            new DecodedString(this.descr).encode().replace("-", "%2D")
         );
     }
 
@@ -78,7 +83,7 @@ public final class NamedDescriptor {
      */
     private static String prefix(final String encoded) {
         try {
-            return encoded.substring(0, encoded.indexOf('-'));
+            return encoded.substring(0, encoded.lastIndexOf('-'));
         } catch (final StringIndexOutOfBoundsException exception) {
             throw new IllegalArgumentException(
                 String.format("Invalid encoded method name: %s", encoded),
@@ -93,6 +98,6 @@ public final class NamedDescriptor {
      * @return The decoded method descriptor
      */
     private static String suffix(final String encoded) {
-        return new EncodedString(encoded.substring(encoded.indexOf('-') + 1)).decode();
+        return new EncodedString(encoded.substring(encoded.lastIndexOf('-') + 1)).decode();
     }
 }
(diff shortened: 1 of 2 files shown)


zaguzovmaksim0-hue (approved):
I checked both separator sides of the encoding. `URLEncoder` leaves `-` unescaped, so changing only `indexOf('-')` to `lastIndexOf('-')` would still be wrong when a descriptor contains a legal dashed internal name such as `LFoo-impl;`. This patch handles that case too by escaping descriptor dashes to `%2D` before concatenation, making the final raw dash unambiguously the separator.

The added cases cover the two relevant ambiguities: a method name with a dash (`box-impl`, `annotation-0`) and a descriptor with a dash (`(LFoo-impl;)V`). Decoding `%2D` restores the original descriptor through `URLDecoder`.

I found no production decode call site outside `NamedDescriptor` itself/tests that would require preserving the old ambiguous representation, and the new encoder/decoder round-trip is internally consistent. No correctness issue found. Approving.

yegor256:
Merged, thanks! Escaping the dash of the descriptor makes the last dash the separator, which is what the split needs.

---

0crat:
@zaguzovmaksim0-hue Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (16) hits-of-code; +4 to give you at least something.  Your running score is [+124](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (16) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5897](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1678, opened 2026-09-10
#1673: keep a trailing group that is no number in the name

### LINKED ISSUE #1673: `NumberedName` refuses a method name with a dash, so a Kotlin `box-impl` cannot be assembled back
**What happens**

`NumberedName` splits a name on the last dash and insists the tail is a number:

```java
private static int suffix(final String encoded) {
    final int result;
    final int index = encoded.lastIndexOf('-');
    if (index == -1) {
        result = 1;
    } else {
        final String number = encoded.substring(index + 1);
        try {
            result = Integer.parseInt(number);
        } catch (final NumberFormatException ex) {
            throw new IllegalArgumentException(
                String.format("Invalid number in name: %s", number),
                ex
            );
        }
    }
    return result;
}
```

A dash is legal in a JVM method name. JVMS 4.2.2 forbids only `.`, `;`, `[`
and `/` in an unqualified name, plus `<` and `>` for methods. Kotlin uses this
for value classes and emits `box-impl`, `unbox-impl`, `constructor-impl`,
`equals-impl`, `toString-impl` and friends.

The two sides of the round trip disagree about such a name. The disassembler
writes it out unchanged, because the number is 1:

```
write side: NumberedName(1, "box-impl").toString() = box-impl
```

and the assembler, reading the same string back in `XmlMethod.name()` through
`.map(NumberedName::new)`, refuses it:

```
read side : new NumberedName("box-impl") -> IllegalArgumentException: Invalid number in name: impl
```

**Impact**

`jeo:disassemble` succeeds on a class with such a method and `jeo:assemble`
then fails on the XMIR it just produced. The message names neither the method
nor the class nor the file, only the fragment `impl`, so it points nowhere near
the cause.

Any Kotlin project with a value class hits this, and so does any JVM language
that puts a dash in a generated method name.

**Evidence**

Both lines above were produced against `target/classes`. `get-x` and
`constructor-impl` fail the same way; `foo` and `foo-2` behave as documented.

**What should happen**

A trailing dash group that is not a number should be treated as part of the
name, the way it is on the way out, so that a name the disassembler emits can
always be read back.

morphqdd:
`NumberedName` split a name on the last dash and insisted the tail was a
number, while the disassembler writes such a name out unchanged. So
`jeo:disassemble` succeeded on a Kotlin value class and `jeo:assemble` then
refused the XMIR it had just produced, with a message naming only the fragment
`impl`.

A trailing group that is not a number is part of the name now, the way it is on
the way out.

Closes #1673

### DIFF
--- a/src/main/java/org/eolang/jeo/representation/NumberedName.java
+++ b/src/main/java/org/eolang/jeo/representation/NumberedName.java
@@ -4,6 +4,8 @@
  */
 package org.eolang.jeo.representation;
 
+import java.util.regex.Pattern;
+
 /**
  * Name representation with optional numeric suffix.
  *
@@ -13,6 +15,15 @@ package org.eolang.jeo.representation;
  * @since 0.9.0
  */
 public final class NumberedName {
+
+    /**
+     * A trailing group that is a number of a name.
+     *
+     * <p>A dash is legal in a JVM method name and Kotlin emits "box-impl" for
+     * a value class, so a trailing group that is not a number belongs to the
+     * name. The digits are bounded to keep the number inside an int.</p>
+     */
+    private static final Pattern NUMBER = Pattern.compile("[1-9][0-9]{0,8}");
 
     /**
      * Number of the name.
@@ -72,18 +83,10 @@ public final class NumberedName {
     private static int suffix(final String encoded) {
         final int result;
         final int index = encoded.lastIndexOf('-');
-        if (index == -1) {
+        if (index == -1 || !NumberedName.NUMBER.matcher(encoded.substring(index + 1)).matches()) {
             result = 1;
         } else {
-            final String number = encoded.substring(index + 1);
-            try {
-                result = Integer.parseInt(number);
-            } catch (final NumberFormatException ex) {
-                throw new IllegalArgumentException(
-                    String.format("Invalid number in name: %s", number),
-                    ex
-                );
-            }
+            result = Integer.parseInt(encoded.substring(index + 1));
         }
         return result;
     }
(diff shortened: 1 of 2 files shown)


zaguzovmaksim0-hue (changes requested):
The non-numeric-tail idea is correct, but the new `NUMBER = [1-9][0-9]{0,8}` changes existing valid numeric suffix semantics beyond #1673.

For example, current `master` accepts `foo-2147483647`: `Integer.parseInt("2147483647")` succeeds, so the decoded base is `foo` and the suffix is a valid positive `int`. This PR's regex rejects every 10-digit suffix, so the same encoded value becomes the literal base name `foo-2147483647` with implicit number 1. Leading-zero forms such as `foo-02` also used to decode numerically (and normalize to `foo-2` on output) but become literal names here.

The distinction should be "numeric tail vs non-numeric tail", not "1-9 digits of a chosen width". A safer shape is to first check whether the tail consists of digits; if it does, keep `Integer.parseInt()` (preserving overflow/zero behavior), and only treat a genuinely non-numeric tail such as `impl` or `x` as part of the base name. Please add regression cases for a 10-digit in-range suffix (`2147483647`) and, ideally, the prior leading-zero behavior.

yegor256:
Merged, thanks!

---

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (39) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5909](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@zaguzovmaksim0-hue Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted.  Your running score is [+363](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1679, opened 2026-09-10
#1672: close the stream even when logging a file fails

### LINKED ISSUE #1672: `Disassembler` loses the summary line when logging a file fails, because the stream is not closed
**What happens**

`Summary` reports the totals from the stream's close handler:

```java
return this.original.apply(representations)
    .peek(rep -> counter.incrementAndGet())
    .onClose(
        () -> Logger.info(
            this,
            "Total %d files were %s in %[ms]s",
            counter.get(),
            this.participle,
            System.currentTimeMillis() - start
        )
    );
```

and `Disassembler` closes it as a plain statement after consuming it:

```java
stream.forEach(this::log);
stream.close();
```

`log` throws when the size of a file cannot be read:

```java
} catch (final IOException exception) {
    throw new IllegalStateException(
        String.format("Failed to get size of '%s'", disassembled),
        exception
    );
}
```

so that exception skips `stream.close()` and the summary line is never
printed. The build fails without the line that says how many files had been
done and how long they took, which is the one number worth having when a run
dies part way through. `Assembler` has the same pair of statements.

**What should happen**

The stream should be closed in a try-with-resources, so that the summary is
logged on the failing path as well.

morphqdd:
The stream was closed by a plain statement after it was consumed, and `log`
throws when the size of a file cannot be read. The summary line lives in the
close handler, so that failure took the count of what was disassembled with it.

The stream is closed by try-with-resources now, so the summary is printed
either way.

Closes #1672

### DIFF
--- a/src/main/java/org/eolang/jeo/Disassembler.java
+++ b/src/main/java/org/eolang/jeo/Disassembler.java
@@ -122,15 +122,15 @@ public final class Disassembler {
         final String process = "Disassembling";
         final String disassembled = "disassembled";
         final Counter counter = new Counter(this.classes.total());
-        final Stream<Path> stream = new Summary(
+        try (Stream<Path> stream = new Summary(
             process,
             disassembled,
             this.classes.toString(),
             this.target,
             new ParallelTranslator(path -> this.disassemble(path, counter), this.threads)
-        ).apply(this.classes.all());
-        stream.forEach(this::log);
-        stream.close();
+        ).apply(this.classes.all())) {
+            stream.forEach(this::log);
+        }
     }
 
     /**


zaguzovmaksim0-hue (approved):
I checked the failure path against `Stream` close semantics. `Summary.apply(...)` attaches its totals log via `onClose`; with the old plain `stream.close()` statement, any exception from `stream.forEach(this::log)` bypasses that callback. Wrapping exactly the same stream in try-with-resources guarantees `close()` runs during unwinding, so the summary callback executes even when `log` throws.

The transformation pipeline and iteration order are unchanged, and the full CI matrix including the long integration test is green. This fixes the resource/lifecycle bug described by #1672 without swallowing the original logging exception. Approving.

yegor256:
Merged, thanks!

---

0crat:
@zaguzovmaksim0-hue Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (8) hits-of-code; +4 to give you at least something.  Your running score is [+60](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (8) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5921](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1680, opened 2026-09-10
#1669: do not report a cached file as disassembled

### LINKED ISSUE #1669: A file skipped by `Caching` is still reported as disassembled by `Logging`
**What happens**

`Disassembler` wraps the caching decorator inside the logging one:

```java
final Transformation trans = new Logging(
    "Disassembling",
    "disassembled",
    new Caching(
        new Informative(
            new Disassembling(this.classes.root(), this.target, path, this.params)
        )
    ),
    this.debug,
    counter
);
```

`Caching.tryTransform` decides that there is nothing to do and says so:

```java
Logger.info(
    this,
    "The file '%s' is already transformed to '%s'. Skipping.",
    this.source(),
    target
);
result = Files.readAllBytes(target);
```

but `Logging.transform` has already started its clock and reports the same
file as processed once the call returns:

```
[INFO] The file 'target/classes/Foo.class' is already transformed to '...Foo.xmir'. Skipping.
[INFO] 12/300 Foo.xmir (4Kb) disassembled in 1ms
```

So an incremental build prints two contradictory lines for every file, and the
`N/M` progress counter counts files that were only read back from disk. There
is no way to tell from the log how much of a build was actual work.

**What should happen**

A file that was served from the cache should not be reported as disassembled.
Either the logging decorator should sit inside the caching one, or the cached
path should say plainly that it was reused.

morphqdd:
The logging decorator sat outside the caching one, so a file served from the
cache printed both "already transformed ... Skipping" and "N/M ... disassembled
in 1ms", and the progress counter counted files that were only read back from
disk.

Caching is the outer decorator now, so a cached file says only that it was
reused and the counter follows the actual work.

Closes #1669

### DIFF
--- a/src/main/java/org/eolang/jeo/Disassembler.java
+++ b/src/main/java/org/eolang/jeo/Disassembler.java
@@ -140,16 +140,16 @@ public final class Disassembler {
      * @return Path to the disassembled XMIR file
      */
     private Path disassemble(final Path path, final Counter counter) {
-        final Transformation trans = new Logging(
-            "Disassembling",
-            "disassembled",
-            new Caching(
+        final Transformation trans = new Caching(
+            new Logging(
+                "Disassembling",
+                "disassembled",
                 new Informative(
                     new Disassembling(this.classes.root(), this.target, path, this.params)
-                )
-            ),
-            this.debug,
-            counter
+                ),
+                this.debug,
+                counter
+            )
         );
         trans.transform();
         return trans.target();


zaguzovmaksim0-hue (approved):
I checked the decorator order against both implementations. `Caching.transform()` calls `origin.transform()` only on a cache miss; on a hit it reads the target bytes directly. With `Caching` outside `Logging`, a cache hit therefore never enters `Logging.transform()`, so it neither starts the processing timer nor advances the `Counter`. On a miss, the inner `Logging(Informative(Disassembling))` still sees the same source/target because both decorators delegate `source()` and `target()` to their origin.

That gives exactly the requested behavior: cached files keep the existing "already transformed ... Skipping" message, while only real disassembly work contributes the `N/M ... disassembled` progress line. No semantic change to cache freshness or output writing. Approving; the remaining long CI job is unrelated to this decorator ordering and the rest of the matrix is green.

yegor256:
Merged, thanks!

---

0crat:
@zaguzovmaksim0-hue Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (16) hits-of-code; +4 to give you at least something.  Your running score is [+56](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (16) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+5933](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1681, opened 2026-09-10
#1666: look at the disabled flag first

### LINKED ISSUE #1666: `DisassembleMojo` initializes the classloader before checking `jeo.disassemble.disabled`
**What happens**

`DisassembleMojo.execute` sets up the classloader before it looks at the flag
that is supposed to turn the goal off:

```java
new PluginStartup(this.project, src).init();
if (this.disabled) {
    Logger.info(this, "Disassemble mojo is disabled, skipping");
} else {
```

`PluginStartup` is not cheap and not neutral. Its constructor calls
`project.getRuntimeClasspathElements()`, `getCompileClasspathElements()` and
`getTestClasspathElements()`, each of which throws
`DependencyResolutionRequiredException` when the classpath is not resolved
yet, and `init()` replaces the context classloader of the current thread with
a `JeoClassLoader` that stays attached for the rest of the build.

So with `jeo.disassemble.disabled=true` the goal still resolves three
classpaths and swaps the classloader, and can fail the build with a dependency
resolution error, which is the opposite of what a user setting `disabled`
expects. `AssembleMojo` has the same shape.

**What should happen**

The `disabled` check should come first, so that a disabled goal does nothing
at all.

morphqdd:
`PluginStartup` ran before the `disabled` flag was read. It resolves three
classpaths, each of which can throw when the classpath is not resolved yet, and
replaces the context classloader for the rest of the build. So
`jeo.disassemble.disabled=true` still did all of that and could fail the build,
which is the opposite of what the flag says.

The flag is read first now, and a disabled goal does nothing.

Closes #1666

### DIFF
--- a/src/main/java/org/eolang/jeo/DisassembleMojo.java
+++ b/src/main/java/org/eolang/jeo/DisassembleMojo.java
@@ -250,11 +250,11 @@ public final class DisassembleMojo extends AbstractMojo {
         this.checkedThreads();
         final Path src = new MavenPath(this.sourcesDir).resolve();
         final Path out = new MavenPath(this.outputDir).resolve();
-        try {
-            new PluginStartup(this.project, src).init();
-            if (this.disabled) {
-                Logger.info(this, "Disassemble mojo is disabled, skipping");
-            } else {
+        if (this.disabled) {
+            Logger.info(this, "Disassemble mojo is disabled, skipping");
+        } else {
+            try {
+                new PluginStartup(this.project, src).init();
                 final boolean listings = !this.omitListings;
                 final boolean comments = !this.omitComments;
                 Logger.info(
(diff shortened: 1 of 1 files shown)


zaguzovmaksim0-hue (changes requested):
The classloader side effect is fixed, but the disabled flag still is not the first gate as #1666 requests. `execute()` still calls `checkedThreads()` and resolves `sourcesDir` / `outputDir` before entering `if (this.disabled)`.

There is a concrete behavioral counterexample: with `jeo.disassemble.disabled=true` and `jeo.disassemble.threads=-1`, this PR still throws `MojoExecutionException` from `checkedThreads()` instead of logging "disabled, skipping" and doing nothing. That contradicts both the issue's stated expected behavior ("the disabled check should come first") and the PR description ("a disabled goal does nothing").

Please make `disabled` the first branch in `execute()` and move thread validation/path resolution into the enabled branch. A regression test combining `disabled=true` with an otherwise-invalid `threads=-1` would pin the intended no-op semantics.

yegor256:
Thanks for this. #1686 was merged for the same issue: it moves the check into `AssembleMojo` as well and adds an integration test that pins the behaviour, so this one is now redundant. Nothing wrong with the change itself.

---

morphqdd:
Not needed anymore, #1686 landed the same fix for #1666.

0crat:
@zaguzovmaksim0-hue Thanks for the review! You've earned +4 points for this: +12 as a basis; -8 for absolutely no comments posted; -4 for too few (22) hits-of-code; +4 to give you at least something.  Your running score is [+367](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1682, opened 2026-09-10
#1600: invalidate the cache when the options change

### LINKED ISSUE #1600: `Caching` ignores the disassembly options, so changing `mode` reuses the old XMIR
## What happens

`Caching` decides that a file is up to date by comparing two modification times only:

https://github.com/objectionary/jeo-maven-plugin/blob/master/src/main/java/org/eolang/jeo/Caching.java

```java
private boolean alreadyTransformed() throws IOException {
    final Path source = this.source();
    final Path target = this.target();
    return Files.exists(target)
        && Files.exists(source)
        && Files.getLastModifiedTime(target).compareTo(Files.getLastModifiedTime(source)) >= 0;
}
```

Nothing in there knows about the options the XMIR was produced with, so changing them and
running again without `mvn clean` keeps the old file:

```java
run(from, out, clazz, "short");   // 82408 bytes written
run(from, out, clazz, "debug");   // 82408 bytes, untouched
```

against the sizes the two modes really produce:

```
short fresh: 82391 bytes
debug fresh: 100471 bytes
```

So the second run says "The file ... is already transformed ... Skipping" and the debug
XMIR, with the line numbers and the local variables in it, is never written. The same
holds for `omitComments`, `omitListings`, `prettyXmir` and `xmir.modifiers`, and for a
plugin upgrade that changes the format.

## What should happen

The cache should be invalidated when the options that produced the file change, for
example by keeping them (and the plugin version) next to the output and comparing them
before the skip.

morphqdd:
`Caching` compared modification times only, so changing `mode`, `omitComments`,
`omitListings`, `prettyXmir` or `xmir.modifiers` and running again without
`clean` kept the XMIR of the previous options and said it was skipping.

The options of the run are written next to the target and compared before the
skip. `Format.fingerprint()` spells them; the assemble side passes nothing and
behaves as before.

Draft: it touches `Disassembler.java`, which #1680 also changes.

Closes #1600


morphqdd:
Rebased and the qulice violation is fixed: the fingerprint moved out of `Format` into its own class, so `TooManyMethods` no longer fires. CI is green. @yegor256, please take a look.

PR #1683, opened 2026-09-10
#1670: Read file size outside counter lock

### LINKED ISSUE #1670: `Logging` reads the file size while holding the counter lock, serializing the parallel phase
**What happens**

`Logging.logEndWithSize` takes a lock on the shared counter and then does file
I/O while holding it:

```java
private void logEndWithSize(final Path source, final Path after, final long time) {
    synchronized (this.counter) {
        if (this.debug) {
            Logger.info(
                this,
                "%s %[file]s %s to %[file]s (%[size]s) in %[ms]s",
                this.counter.next(),
                source,
                this.participle,
                after,
                Logging.size(after),
                time
            );
```

`Logging.size` calls `Files.exists` and `Files.size`, so every worker thread
that finishes a class queues up behind two filesystem calls made by whichever
thread holds the lock. One `Logging` instance is created per file but they all
share the same `Counter`, so the lock is global to the run.

The whole point of `ParallelTranslator` is to use every core, and this puts a
serial section with disk access at the end of each unit of work.

**What should happen**

The size should be read before the lock is taken, so that only the counter
increment and the log line are serialized.

zaguzovmaksim0-hue:
Closes #1670.

`Logging.logEndWithSize()` used to call `Files.exists()` / `Files.size()` through `Logging.size(after)` while synchronized on the shared `Counter`. Every per-file `Logging` instance in the parallel run shares that counter, so filesystem I/O became part of the global serial section.

This computes the target size once before taking the counter lock and reuses the value in either logging branch. The synchronized block is now limited to advancing the shared counter and emitting the ordered progress line.

The log format, counter behavior, error handling, and measured file are unchanged; only the lock scope is reduced. `git diff --check` is clean.

### DIFF
--- a/src/main/java/org/eolang/jeo/Logging.java
+++ b/src/main/java/org/eolang/jeo/Logging.java
@@ -114,6 +114,7 @@ public final class Logging implements Transformation {
      * @param time Time spent in milliseconds
      */
     private void logEndWithSize(final Path source, final Path after, final long time) {
+        final long size = Logging.size(after);
         synchronized (this.counter) {
             if (this.debug) {
                 Logger.info(
@@ -123,7 +124,7 @@ public final class Logging implements Transformation {
                     source,
                     this.participle,
                     after,
-                    Logging.size(after),
+                    size,
                     time
                 );
             } else {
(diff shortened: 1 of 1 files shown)


zaguzovmaksim0-hue:
Closing this PR to comply with the maintainer policy: issue #1670 is currently unlabeled. I will not submit work against unlabeled issues; if the issue is accepted/labeled later, the work can be reconsidered then.

PR #1684, opened 2026-09-10
#1668: Traverse inputs once per run

### LINKED ISSUE #1668: `Classes.total()` walks the directory a second time and logs `Found N files` twice
**What happens**

`Classes.total()` is implemented as a full traversal:

```java
@Override
public long total() {
    return this.all().count();
}
```

and `Disassembler.disassemble()` asks for both, one after the other:

```java
final Counter counter = new Counter(this.classes.total());
final Stream<Path> stream = new Summary(
    ...
).apply(this.classes.all());
```

So `Files.walk` runs twice over `target/classes` on every disassemble, once
only to learn how many files there are.

With `FilteredClasses` in between, which is what the mojo builds, the second
pass is not just a walk. `all()` collects the whole list, applies every glob
matcher to every path, and logs:

```java
this.logger.accept(
    String.format("Found %d files in %s using %s", size, this.original, this.filter)
);
```

so the same `Found N files in ...` line is printed twice per run, once from
the counting call and once from the real one, and the globs are matched twice.

`Assembler` does the same with `XmirFiles`.

**What should happen**

The list should be walked once and reused for both the count and the work.

zaguzovmaksim0-hue:
Closes #1668.

Both top-level pipelines counted their inputs with a full traversal and then requested the same inputs again for processing:

- `Disassembler`: `classes.total()` -> `classes.all().count()`, then `classes.all()` again.
- `Assembler`: `files.total()` -> `files.all().count()`, then `files.all()` again.

This materializes each input stream once, initializes the progress `Counter` from the resulting list size, and feeds the same list into `Summary`/`ParallelTranslator`. For `FilteredClasses`, this also means its filtering and `Found N files ...` log run once instead of twice.

No ordering, filtering, transformation, or progress-total semantics change. `git diff --check` is clean; the repository CI/JMH jobs can measure the traversal reduction on the supported build environment.

### DIFF
--- a/src/main/java/org/eolang/jeo/Assembler.java
+++ b/src/main/java/org/eolang/jeo/Assembler.java
@@ -8,6 +8,8 @@ import com.jcabi.log.Logger;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.List;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import org.eolang.jeo.representation.Counter;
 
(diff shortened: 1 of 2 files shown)


zaguzovmaksim0-hue:
Closing this PR to comply with the maintainer policy: issue #1668 is currently unlabeled. I will not submit work against unlabeled issues; if the issue is accepted/labeled later, the work can be reconsidered then.

PR #1685, opened 2026-09-10
#1671: Index labels once per flow

### LINKED ISSUE #1671: `InstructionsFlow.index()` scans every instruction, making the max-stack computation quadratic
**What happens**

`InstructionsFlow.index` finds a label by scanning the whole instruction list:

```java
private int index(final BytecodeLabel label) {
    for (int index = 0; index < this.instructions.size(); ++index) {
        if (this.instructions.get(index).equals(label)) {
            return index;
        }
    }
    throw new IllegalStateException(String.format("Label %s not found", label));
}
```

It is called from the main loop for every branch and jump, and from
`suitableBlocks`, which runs for every instruction that is not a jump and
scans three times per try-catch block:

```java
return this.blocks.stream()
    .map(BytecodeTryCatchBlock.class::cast)
    .filter(block -> this.index(block.startLabel()) <= instruction)
    .filter(block -> this.index(block.endLabel()) >= instruction)
    .map(block -> this.index(block.handlerLabel()))
    .collect(Collectors.toList());
```

So computing max stack for a method costs on the order of instructions times
instructions times blocks, and a worklist entry can revisit the same
instruction more than once. A generated method with a few thousand
instructions and a handful of try-catch blocks pays for millions of list
comparisons where a lookup table would answer in constant time.

**What should happen**

The label to index mapping should be built once for the method and looked up,
instead of being searched linearly on each use.

zaguzovmaksim0-hue:
Closes #1671.

`InstructionsFlow.index()` previously scanned the complete instruction list on every branch lookup and on each try/catch boundary lookup. The same labels are immutable for the lifetime of one flow, so these repeated linear searches multiply inside the worklist analysis.

This builds a `BytecodeLabel -> instruction index` map once in the constructor and makes `index()` a constant-time lookup. `BytecodeLabel` already has `@EqualsAndHashCode`, so its existing equality semantics are preserved as map-key semantics. `putIfAbsent` deliberately keeps the first equal label, matching the previous forward linear search if duplicate-equivalent labels ever appear.

Missing labels still throw the same `IllegalStateException("Label ... not found")`; flow traversal and max-value logic are unchanged. `git diff --check` is clean. The repository's max-stack/max-locals and integration CI will exercise the functional behavior on the supported build environment.

### DIFF
--- a/src/main/java/org/eolang/jeo/representation/bytecode/InstructionsFlow.java
+++ b/src/main/java/org/eolang/jeo/representation/bytecode/InstructionsFlow.java
@@ -9,6 +9,7 @@ import java.util.ArrayList;
 import java.util.Deque;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.function.Function;
 import java.util.stream.Collectors;
@@ -33,6 +34,11 @@ public final class InstructionsFlow<T extends InstructionsFlow.Reducible<T>> {
     private final List<BytecodeTryCatchBlock> blocks;
 
     /**
+     * Instruction indexes of labels.
+     */
+    private final Map<BytecodeLabel, Integer> labels;
+
+    /**
      * Constructor.
      * @param instr Instructions.
      * @param catches Try-catch blocks.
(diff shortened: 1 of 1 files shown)


zaguzovmaksim0-hue:
Closing this PR to comply with the maintainer policy: issue #1671 is currently unlabeled. I will not submit work against unlabeled issues; if the issue is accepted/labeled later, the work can be reconsidered then.

PR #1686, opened 2026-09-10
#1666: Exit disabled goals before initialization

### LINKED ISSUE #1666: `DisassembleMojo` initializes the classloader before checking `jeo.disassemble.disabled`

zaguzovmaksim0-hue:
Closes #1666.

Both goals now return immediately when their `disabled` flag is true, before validating `threads`, resolving source/output paths, or initializing any classloader. This matches the existing parameter documentation that a disabled goal skips all processing and exits immediately.

The new Maven Invoker test configures both `disassemble` and `assemble` with:

- `disabled=true`
- deliberately invalid `threads=-1`

Before this change the build fails in `checkedThreads()` before reaching the disabled branch. With the early return both executions log their disabled message and the build succeeds. For `disassemble`, this also guarantees `PluginStartup.init()` is not reached while disabled.

I could not run the Maven Invoker suite locally because Maven is not installed in this Android/Termux environment; the fixture POM is XML-validated locally and repository CI is the integration validation source.

### DIFF
--- a/src/it/disabled/invoker.properties
+++ b/src/it/disabled/invoker.properties
@@ -1,0 +1,4 @@
+# SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+# SPDX-License-Identifier: MIT
+invoker.goals=clean process-classes
+invoker.java.version = 17+
--- a/src/it/disabled/pom.xml
+++ b/src/it/disabled/pom.xml
@@ -1,0 +1,51 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+* SPDX-FileCopyrightText: Copyright (c) 2016-2026 Objectionary.com
+* SPDX-License-Identifier: MIT
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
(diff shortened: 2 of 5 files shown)


yegor256:
Merged, thanks! The early return in both mojos plus the `disabled` integration test is exactly what #1666 asked for.

---

0crat:
@zaguzovmaksim0-hue Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too many hits-of-code (177 >= 100). Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+351](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1467, opened 2026-04-09
build(#1463): update lints version to 0.1.0 and remove disabled tests

volodya-lombrozo:
This PR updates the `lints` version in `pom.xml` and removes disabled tests in `XmirFilesTest.java`.

Closes #1463

## Summary by CodeRabbit

* **Chores**
  * Updated the Maven linting/validation dependency to a newer version.

* **Bug Fixes**
  * Improved validation by expanding the set of ignored defect variants during XMIR processing to reduce false failures.

* **Tests**
  * Re-enabled two previously disabled tests to restore full test coverage.

PR #1687, opened 2026-09-10
#1599: Let debug XMIR pass validation

### LINKED ISSUE #1599: XMIR produced in the default `debug` mode fails the plugin's own verification
## What happens

The XMIR that `disassemble` produces in `debug` mode does not pass the verification that
the same plugin offers:

```java
final String xmir = new BytecodeRepresentation(new Bytecode(bytes))
    .toEO(new Format(Format.MODE, "debug"));
Files.write(tmp, xmir.getBytes(StandardCharsets.UTF_8));
new JcabiXmlDoc(tmp).validate();
```

```
short: VALID
debug: XMIR is incorrect (102 errors): [j$org.j$eolang.j$jeo.j$Caching line-is-absent/S ERROR]
            The @line attribute is absent at number number
```

That was `org/eolang/jeo/Caching.class`, one of the plugin's own classes. The same happens
for every class with code that I tried; only a class with no methods passes.

`debug` is the default of the goal:

https://github.com/objectionary/jeo-maven-plugin/blob/master/src/main/java/org/eolang/jeo/DisassembleMojo.java

```java
@Parameter(
    property = "jeo.disassemble.mode",
    defaultValue = "debug"
)
private String mode;
```

and the verification is the documented option next to it:

```java
@Parameter(
    property = "jeo.disassemble.xmir.verification",
    defaultValue = "false"
)
private boolean xmirVerification;
```

so `-Djeo.disassemble.xmir.verification=true` fails the build on any real project, with
errors about the plugin's own output. `-Djeo.assemble.xmir.verification=true` reads the
same files, so it cannot be used either.

The lint is `line-is-absent`, which wants a `@line` attribute on the objects that the
debug mode adds for the line numbers.

## What should happen

The XMIR of the default mode should pass the verification the plugin ships, so that
turning the option on is a way to catch a broken file rather than a way to fail every
build.

zaguzovmaksim0-hue:
Closes #1599.

JEO's validation already suppresses EO-source lints that are not valid invariants for its synthetic bytecode XMIR. `line-is-absent` is the same kind of false positive: bytecode does not have an EO source-line contract, and classes may omit `LineNumberTable` entirely. In debug mode JEO additionally materializes JVM line-number structures; requiring every synthetic `<o>` inside those structures to carry an EO `@line` makes the plugin reject its own valid output.

This adds `line-is-absent` (and, through the existing helper, its `/S` variant) to the central validation ignore set. It does not disable structural/XSD verification or unrelated lints.

A regression test now generates XMIR from the real `MethodByte.class` fixture in explicit `debug` mode and runs the same `XmirFiles.verify()` path used by the mojo. The existing short-mode validation tests remain unchanged.

This follows the existing validation design from #1467, where JEO-specific false-positive lints are centralized in `JcabiXmlNode.ignored()` instead of fabricating source metadata in generated XMIR.

Maven is not installed in this Android/Termux environment, so repository CI is the test execution source; `git diff --check` is clean locally.

### DIFF
--- a/src/main/java/org/eolang/jeo/representation/xmir/JcabiXmlNode.java
+++ b/src/main/java/org/eolang/jeo/representation/xmir/JcabiXmlNode.java
@@ -195,7 +195,9 @@ public final class JcabiXmlNode implements XmlNode {
             "sparse-decoration",
             "idempotent-attribute-is-not-first",
             "empty-object",
-            "object-has-data"
+            "object-has-data",
+            // Bytecode has no EO source-line contract (#1599).
+            "line-is-absent"
         );
         return Stream.concat(
             rules.stream(),
--- a/src/main/java/org/eolang/jeo/representation/xmir/JcabiXmlNode.java
+++ b/src/main/java/org/eolang/jeo/representation/xmir/JcabiXmlNode.java
@@ -196,7 +196,6 @@ public final class JcabiXmlNode implements XmlNode {
             "idempotent-attribute-is-not-first",
             "empty-object",
             "object-has-data",
-            // Bytecode has no EO source-line contract (#1599).
             "line-is-absent"
         );
         return Stream.concat(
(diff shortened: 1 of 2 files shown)


yegor256:
CI is green and the change is small, but I don't want to merge it without a maintainer's call: it makes the verification pass by adding `line-is-absent` to the ignore set rather than by making debug-mode XMIR carry the `@line` the lint asks for. That is a policy decision about what JEO's own XMIR is expected to satisfy, not a mechanical fix. Assigning to @yegor256.

---

PR #1689, opened 2026-09-10
#1667: Validate disassembly mode before work

### LINKED ISSUE #1667: `jeo.disassemble.mode` is validated once per class instead of once per run, unlike `threads`
**What happens**

`DisassembleMojo` validates one of its two free-form parameters before it
starts:

```java
@Override
public void execute() throws MojoExecutionException {
    this.checkedThreads();
```

`mode` gets no such treatment. It is passed into `Format` as a plain string
and only reaches its validator far down in the work, once per class, inside
the parallel translation:

```java
.bytecode(DisassembleMode.fromString(fmt.mode()).asmOptions());
```

`DisassembleMode.fromString` throws `IllegalArgumentException` for anything
that is not `short` or `debug`. By that point the run is inside
`ParallelTranslator`, so a single typo in `-Djeo.disassemble.mode=Debag`
surfaces as an `ExecutionException` wrapped in `IllegalStateException`
("Parallel translation failed"), after the plugin has already logged that
disassembling started in mode `Debag` and started N threads.

The message that says which values are allowed is buried at the bottom of that
chain, and the same mistake in `threads` gives a clean one-line Maven error.

**What should happen**

`mode` should be validated in `execute()` next to `checkedThreads()`, so that a
wrong value is a configuration error before any work begins.

zaguzovmaksim0-hue:
Closes #1667.

`DisassembleMojo` now validates `jeo.disassemble.mode` immediately after validating `threads`, before resolving Maven paths, initializing the plugin classloader, logging that disassembly started, or creating parallel workers.

An invalid value from `DisassembleMode.fromString()` is converted to `MojoExecutionException` with the original diagnostic, so a typo such as `Debag` surfaces as a configuration error instead of being wrapped later as a parallel-translation failure.

Added a regression test that intentionally leaves the work paths unconfigured and sets only `mode=Debag`; it can pass only when the mode is rejected before disassembly setup. Reflection is confined to the test helper with the same targeted PMD suppression already used by the project for accessibility alteration.

`git diff --check` is clean. Maven is not installed in this Termux environment, so the repository CI matrix will run the Java tests.

### DIFF
--- a/src/main/java/org/eolang/jeo/DisassembleMojo.java
+++ b/src/main/java/org/eolang/jeo/DisassembleMojo.java
@@ -16,6 +16,7 @@ import org.apache.maven.plugins.annotations.Mojo;
 import org.apache.maven.plugins.annotations.Parameter;
 import org.apache.maven.project.MavenProject;
 import org.cactoos.set.SetOf;
+import org.eolang.jeo.representation.asm.DisassembleMode;
 import org.eolang.jeo.representation.directives.Format;
 
 /**
@@ -248,6 +249,7 @@ public final class DisassembleMojo extends AbstractMojo {
     @Override
     public void execute() throws MojoExecutionException {
         this.checkedThreads();
+        this.checkedMode();
         final Path src = new MavenPath(this.sourcesDir).resolve();
         final Path out = new MavenPath(this.outputDir).resolve();
         try {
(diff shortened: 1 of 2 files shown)


zaguzovmaksim0-hue:
Closing this PR to comply with the maintainer policy: issue #1667 is currently unlabeled. I will not submit work against unlabeled issues; if the issue is accepted/labeled later, the work can be reconsidered then.

PR #1700, opened 2026-09-14
#1692: read the number of locals by the name the old format wrote

### LINKED ISSUE #1692: `XmlFrame` looks for `nlocals`, while the legacy format writes `nlocal`
## What happens

`XmlFrame` reads the number of locals of a legacy frame by a name that the legacy format never
wrote (`src/main/java/org/eolang/jeo/representation/xmir/XmlFrame.java:112-113`):

```java
private Optional<Integer> nlocals() {
    return this.byName("nlocals")
```

used at line 63 as `this.nlocals().orElse(locals.length)`. Its sibling asks for `"nstack"` and
works. `byName` matches with `startsWith` (`XmlFrame.java:126-132`), and the old format wrote
`nlocal-<hash>` and `nstack-<hash>`:

```
$ git show 9db6bef2b^:.../directives/DirectivesFrame.java
new DirectivesValue(this.name("nlocal"), this.nlocal),
new DirectivesValue(this.name("nstack"), this.nstack),
```

`"nlocal-123".startsWith("nlocals")` is false, so the branch never fires and the value silently
falls back to `locals.length`.

The typo arrived with `9db6bef2b`, the commit that added this compatibility reader. The
regression test `XmlFrameTest.parsesFullFrameNode`, whose javadoc says it verifies backward
compatibility with the old frame representation, passes only because it uses `nlocal = 0`, where
the fallback happens to agree. With `nlocal = 2` and an empty list of locals:

```
GOT>>>      BytecodeFrame(type=2, nlocal=0, locals=[], nstack=0, stack=[])
EXPECTED>>> BytecodeFrame(type=2, nlocal=2, locals=[], nstack=0, stack=[])
```

For `F_CHOP` (type 2) the number is the count of chopped locals and must be between 1 and 3, so
0 makes the emitted `StackMapTable` invalid.

The current format is not affected: ASM gives `F_CHOP` a list of N nulls and that round-trips
exactly.

## What should happen

The reader should ask for `nlocal`, the name the old format actually wrote, and the regression
test should use a value where the fallback and the real number differ.

morphqdd:
`XmlFrame` asked for `nlocals`, while the old format wrote `nlocal`, so the branch never fired and the number of locals silently fell back to the length of the list. For `F_CHOP` that number is the count of chopped locals, so an empty list gave 0 and the emitted `StackMapTable` was invalid.

The regression test passed only because it used `nlocal = 0`, where the fallback agrees. It now uses 2.

Closes #1692


yegor256:
Merged, thanks! Good catch that the old assertion used `0`, which is exactly the value the failed lookup fell back to, so the test could not see the mismatch.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (6) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+6109](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1701, opened 2026-09-14
#1691: refuse a parameters table longer than the descriptor

### LINKED ISSUE #1691: `AsmMethodParameters` loops over the `MethodParameters` table and indexes the descriptor array
## What happens

`AsmMethodParameters.bytecode()` iterates over the `MethodParameters` table but reads the
descriptor array with the same index
(`src/main/java/org/eolang/jeo/representation/asm/AsmMethodParameters.java:44-58`):

```java
final List<ParameterNode> params = Optional.ofNullable(this.node.parameters)
    .orElse(Collections.emptyList());
final Type[] types = this.types();
for (int index = 0; index < params.size(); ++index) {
    res.add(new BytecodeMethodParameter(index, ..., types[index]));
}
```

JVMS 4.7.24 says that `parameters_count` is not required to be equal to the number of parameter
descriptors in the method descriptor. When the table is longer, `types[index]` throws.

Reproduced with a class built by ASM whose descriptor is `(I)V` and which calls `visitParameter`
three times, then read back with `new AsmProgram(bytes).bytecode(0)` (flags `0` is the
`mode=debug` path, the one `hone-maven-plugin` uses):

```
java.lang.ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1
  at org.eolang.jeo.representation.asm.AsmMethodParameters.bytecode(AsmMethodParameters.java:53)
  at org.eolang.jeo.representation.asm.AsmMethod.bytecode(AsmMethod.java:58)
  at org.eolang.jeo.representation.asm.AsmClass.methods(AsmClass.java:87)
  at org.eolang.jeo.representation.asm.AsmProgram.bytecode(AsmProgram.java:55)
```

The error is raw, with no mention of the class or the method it came from. In `mode=short` it
cannot happen, because `ClassReader.SKIP_DEBUG` skips the attribute entirely.

This is the same class of defect as #939, #964, #967 and #1538.

## What should happen

The loop should stay inside both arrays, and a table that disagrees with the descriptor should
produce a message naming the class and the method rather than a bare
`ArrayIndexOutOfBoundsException`.

morphqdd:
`AsmMethodParameters.bytecode()` walked the `MethodParameters` table and read the descriptor array with the same index. JVMS 4.7.24 does not require the two to agree, so a longer table threw a bare `ArrayIndexOutOfBoundsException` naming neither the class nor the method.

Now the mismatch is reported with the method, its descriptor and both counts.

Closes #1691


yegor256:
Merged, thanks!

---

PR #1721, opened 2026-09-15
#1720: write the line separator the test expects

### LINKED ISSUE #1720: `PrettyXmlTest` expects the platform line separator, while dom4j always writes `\n`
## What happens

The `test (windows-2022, *)` job fails on every pull request rebased on current `master`:

```
[ERROR] org.eolang.jeo.representation.PrettyXmlTest.formatsXml <<< FAILURE!
java.lang.AssertionError:
We can't format XML output with correct indentation (two spaces)
Expected: "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\r\n\r\n<o>\r\n  <o>\r\n ..."
     but: was "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n\n<o>\n  <o>\n ..."
```

`40b5c69cb` replaced the two `"\n"` literals in `PrettyXmlTest` with `System.lineSeparator()`:

```java
-                    "\n",
+                    System.lineSeparator(),
```

`PrettyXml` formats through dom4j's `OutputFormat.createCompactFormat()`, and dom4j's line
separator is `"\n"`, fixed, not the platform one. So on Windows the expected string carries
`\r\n` while the produced one carries `\n`, and the test cannot pass.

Checked directly against dom4j 2.2.0, the version this project pins, with `line.separator` set to
`\r\n`:

```
separator=\n
output=<?xml version="1.0" encoding="UTF-8"?>\n\n<o>\n  <o>\n    <o>3.14</o>\n  </o>\n</o>\n
```

The same failure reproduces on Linux by running the test with `-Dline.separator=\r\n`.

## What should happen

The test should compare against the separator the formatter really emits. Either the two literals
go back to `"\n"`, or `PrettyXml` should be told which separator to use through
`OutputFormat.setLineSeparator` and the test should assert that one.

morphqdd:
`PrettyXml` formats through dom4j, whose line separator is `\n` and does not follow the platform. The test was switched to `System.lineSeparator()`, so on Windows it expects `\r\n` and cannot pass, and the `test (windows-2022, *)` job is red on every pull request rebased on current master.

Reproduces on Linux too, with `-Dline.separator=\r\n`.

Closes #1720


morphqdd:
Changed the approach: instead of putting the `"\n"` literals back in the test, which the new qulice forbids, `PrettyXml` now tells dom4j which separator to write.

dom4j's `OutputFormat` line separator is `"\n"` fixed, so the formatter ignored the platform while the test asked for it. Setting it from `System.lineSeparator()` makes the two agree on both platforms:

```
sep=\n   -> <?xml version="1.0" encoding="UTF-8"?>\n\n<o>\n  <o>...
sep=\r\n -> <?xml version="1.0" encoding="UTF-8"?>\r\n\r\n<o>\r\n  <o>...
```

The test is left exactly as it is on master.

morphqdd:
Superseded by b1231723, which asserts against a literal newline on master. That fixes the Windows job, so this one is no longer needed and its `setLineSeparator` would now push the output back to CRLF and break that assertion.

PR #1722, opened 2026-09-15
#1711: count the operands of a handle before reading them

### LINKED ISSUE #1711: `XmlHandle` reads five operands by index without checking how many there are
## What happens

`XmlHandle.bytecode()` collects the children into a list and reads five of them by index:

```java
final List<XmlNode> operands = this.node.children().collect(Collectors.toList());
... operands.get(0) ... operands.get(4)
```

There is no size check. A `Φ.jeo.handle` object with fewer than five arguments throws a bare
`IndexOutOfBoundsException` that names neither the handle, nor the method it belongs to, nor which
operand is missing. The `Objects.requireNonNull` calls below guard the wrong thing, because the
list access blows up before them.

Reproduced with a handle object carrying one argument:

```
java.lang.IndexOutOfBoundsException: Index 1 out of bounds for length 1
```

Such XMIR is not hypothetical: `jeo` is a format other tools write and edit, and a handle that
lost an argument to a hand edit or to an XMIR level optimizer arrives here.

## What should happen

The operand count should be checked first, and a malformed handle should raise an
`IllegalStateException` that names the node and says how many operands were expected, the way
`XmlAnnotationValue.type()` and `XmlTryCatchEntry.label()` already do in the same package.

morphqdd:
`XmlHandle.bytecode()` read five operands by index with no size check, so a handle carrying fewer threw a bare `IndexOutOfBoundsException` naming neither the handle nor how many operands were missing.

It now counts them first and says what it found, the way `XmlAnnotationValue.type()` does in the same package.

Closes #1711


yegor256:
Merged, thanks!

---

PR #1723, opened 2026-09-15
#1716: find an attribute by its name, not by its prefix

### LINKED ISSUE #1716: Both `XmlNode` implementations find an attribute by prefix, while they find a child by name
## What happens

Both implementations of `XmlNode` look up an attribute by prefix:

```java
// NativeXmlNode
.filter(attr -> attr.getNodeName().startsWith(name))
// JcabiXmlNode
.filter(attr -> attr.getKey().startsWith(name))
```

while the sibling lookup for children in the very same classes, `optchild(name)`, correctly uses
`equals`. So `attribute("base")` returns the value of any attribute whose name merely begins with
`base`, and `attribute("name")` of any attribute beginning with `name`, with no error.

A lot rides on this: `XmlJeoObject.named()` and `name()`, `XmlClass.name()`,
`XmlClosedObject.base()` and the `φ` test in `XmlDelegateObject.base()` all go through it.

Today's writer happens to emit only `base`, `name` and `as`, so nothing collides. But XMIR that has
been through the EO toolchain or a downstream plugin carries extra attributes, and any one of them
starting with `base`, `name` or `as` hijacks the lookup and the class is assembled from the wrong
value instead of failing.

## What should happen

The attribute lookup should compare with `equals`, the way `optchild` in the same class does. The
mismatch is duplicated in both implementations, so both need the change.

morphqdd:
Both `XmlNode` implementations matched an attribute with `startsWith`, while the sibling lookup for children in the same classes uses `equals`. So asking for `name` on a node that carries only `names` answered with the wrong value instead of nothing, and a lot rides on that lookup: `XmlJeoObject.name()`, `XmlClass.name()`, `XmlClosedObject.base()` and the `φ` test in `XmlDelegateObject.base()`.

Both now compare with `equals`. The whole suite of 992 tests stays green.

Closes #1716


yegor256:
Merged, thanks! Both node implementations now agree with how children are looked up, and the full integration suite stayed green, which is the evidence I wanted before tightening a matcher this widely used.

---

0crat:
@morphqdd Thanks for the contribution! You've earned +12 points for this: +16 as a basis; -4 for too few (22) hits-of-code. Please, [keep](https://www.yegor256.com/2018/03/06/speed-vs-quality.html) them coming. Your running score is [+6121](https://www.eolang.org/zerocracy/objectionary-vitals.html); don't forget to check your [Zerocracy account](https://www.zerocracy.com/) too).

PR #1724, opened 2026-09-15
#1712: check the values of an annotation property before reading them

### LINKED ISSUE #1712: `XmlAnnotationValue` indexes its values unguarded, while the check above it guards the same list
## What happens

`XmlAnnotationValue.type()` checks the list before it reads it:

```java
if (collect.isEmpty()) {
    throw new IllegalStateException(...);
}
```

Two methods later, `bytecode()` reads `params.get(0)`, `params.get(1)` and `params.get(2)` with no
equivalent check. An `annotation-property` whose type is `PLAIN`, `ENUM` or `ANNOTATION` but which
is missing its value children throws a bare `IndexOutOfBoundsException`:

```
java.lang.IndexOutOfBoundsException: Index 0 out of bounds for length 0
```

`XmlMethod.bytecode()` catches `IndexOutOfBoundsException` and rewraps it as "most probably due to
broken XMIR", which names the method but never the annotation or the property, so there is nothing
to look for in the file.

## What should happen

The arity should be validated per branch (`PLAIN` needs two children, `ENUM` three, `ANNOTATION` at
least two) and a missing child should raise an `IllegalStateException` naming the property node and
the expected count, the same way `type()` does a few lines above.

morphqdd:
`type()` checks the list before it reads it, and two methods later `bytecode()` read `params.get(0)`, `get(1)` and `get(2)` with no equivalent check. An `annotation-property` missing its value children threw a bare `IndexOutOfBoundsException`, which `XmlMethod.bytecode()` rewraps as "most probably due to broken XMIR", naming the method but never the annotation or the property.

Each branch now validates its own arity and says which node, which type and how many values were expected.

Closes #1712


yegor256:
Merged, thanks!

---

PR #1727, opened 2026-09-15
#1717: refuse an inner class that has no access mask

### LINKED ISSUE #1717: `XmlAttribute` turns a missing `InnerClass` access mask into package private
## What happens

`XmlAttribute` reads the four operands of an `InnerClass` entry like this:

```java
this.operand(0).orElse(null),
this.operand(1).orElse(null),
this.operand(2).orElse(null),
this.operand(3).orElse(0)
```

The first three use `null`, which is the right encoding for "genuinely absent", since an anonymous
inner class really has no name and no outer name. The fourth, the access mask, defaults to `0`.

`0` is not "absent". It is a valid access mask, the one that means package private with no other
modifier. So XMIR whose `inner-class` object has lost its fourth child assembles into a class file
carrying a package private `InnerClasses` entry, instead of failing. The corruption is invisible
until somebody disassembles the emitted `.class` and notices the modifiers changed.

## What should happen

A missing access child should raise an `IllegalStateException` naming the node, the way
`XmlUnknownAttribute` does for its own missing children, rather than inventing a plausible looking
access mask.

morphqdd:
The first three operands of an `InnerClass` entry default to `null`, which is the encoding for genuinely absent. The fourth, the access mask, defaulted to `0`, and `0` is not absent: it is a valid mask meaning package private with no other modifier.

So XMIR whose `inner-class` object had lost its fourth child assembled into a class file carrying a package private `InnerClasses` entry instead of failing, and the corruption was invisible until somebody disassembled the emitted class and noticed the modifiers had changed.

A missing access child is now refused with a message naming the node.

Closes #1717


yegor256:
Merged, thanks!

---

PR #1728, opened 2026-09-15
#1714: skip the first child only when it is the delegate

### LINKED ISSUE #1714: `XmlDelegateObject.children` drops the first child even when it is not the delegate
## What happens

`XmlDelegateObject.base()` is careful. It treats the first child as the delegate only when that
child carries `name="φ"`, and otherwise answers with nothing:

```java
if (first.attribute("name").map("φ"::equals).orElse(false)) { ... }
return Optional.empty();
```

`children()` and `child(int)` are not careful. They drop the first child unconditionally:

```java
return all.subList(1, all.size());
```

So for an object written in the closed form, `<o base="Φ.jeo.trycatch" name="t0">`, which is the
form `DirectivesClosedObject` emits, `base()` correctly says "this is not a delegate object" while
`children()` silently eats a real argument. Every reader that goes through `XmlJeoObject` then
works on an off by one child list: `XmlInstruction.operands()`, `XmlHandle`, `XmlTryCatchEntry`,
`XmlLocalVariable`, `XmlFrame`.

Checked with a `<o base="Φ.jeo.trycatch" name="t0">` holding four children: `base()` answers empty
while `children()` returns three of the four.

`XmlAbstractObject` has the same split between a checked `base()` and an unchecked `children()`.

## What should happen

`children()` and `child(int)` should skip the first child only when `base()` actually recognised it
as the `φ` delegate, so a mis-shaped object cannot be read as a valid one with an argument quietly
missing.

morphqdd:
`base()` is careful: it treats the first child as the delegate only when that child carries `name="φ"`. `children()` and `child(int)` were not, and dropped the first child unconditionally.

So for an object written in the closed form, which is the form `DirectivesClosedObject` emits, `base()` correctly answered "this is not a delegate object" while `children()` silently ate a real argument, and every reader that goes through `XmlJeoObject` worked on an off by one list: `XmlInstruction.operands()`, `XmlHandle`, `XmlTryCatchEntry`, `XmlLocalVariable`, `XmlFrame`.

The first child is now skipped only when `base()` actually recognised it. All 994 tests stay green.

Closes #1714


yegor256:
Merged, thanks!

---

PR #1729, opened 2026-09-16
#1718: refuse to assemble more than one class

### LINKED ISSUE #1718: `BytecodeObject.bytecode` assembles only the first class and drops the rest silently
## What happens

`BytecodeObject` holds a list of classes. The constructor is public and takes that list, and
`equals` and `hashCode` compare all of it. But the assembly path reads only the first one:

```java
public Bytecode bytecode() {
    ... this.top() ...
}
private BytecodeClass top() {
    return this.classes.get(0);
}
```

Everything after the first class is dropped without a word. The reading side mirrors it:
`XmlObject.top()` takes `this.root.child("o")`, and `optchild` stops at the first match, so a second
top level class in the XMIR is dropped before it is ever parsed.

Both goals go through `BytecodeObject`, and it is a public class that `opeo`, `hone` and `sodg`
build directly, so a consumer that produces a multi class object gets one `.class` file back and no
diagnostic at all.

## What should happen

Either every class in the list should be assembled, or a `BytecodeObject` carrying more than one
class should be refused with a message that says so. Silently keeping the first is the one answer
that loses data without telling anybody.

(The empty list case is #1709 and is a different defect in the same method.)

morphqdd:
`BytecodeObject.bytecode` assembled `classes.get(0)` and dropped everything else without a word. A `.class` file holds one top level class, so the list cannot be assembled as it is; now it says so instead of losing the rest silently.

Closes #1718


morphqdd:
@yegor256, please take a look when you have a moment.

PR #1730, opened 2026-09-16
#1719: name the member that is missing

### LINKED ISSUE #1719: `XmlEnclosingMethod` and `XmlSourceFile` report every missing child as the first one
## What happens

`XmlEnclosingMethod` reads three different things through one helper:

```java
private String string(final int index) {
    return this.node.children()...
        .orElseThrow(() -> new IllegalStateException("Enclosing method owner is absent"));
}
```

`owner()`, `method()` and `descriptor()` all call it, so a missing descriptor is reported as a
missing owner. The message also carries neither the class being assembled nor the node, so there is
nothing to grep for in a build log.

`XmlSourceFile` has the same shape: `source()` and `debug()` each throw
`new IllegalStateException("Source file node is absent")`, so a missing debug extension is reported
as a missing source file name.

## What should happen

Each message should name the member that is actually missing and the node it was expected in, the
way `XmlUnknownAttribute` does two files over with `"Attribute 'type' is missing in `%s`"`.

morphqdd:
`XmlEnclosingMethod` reported a missing method name or descriptor as a missing owner, and `XmlSourceFile` reported a missing debug extension as a missing source file. Both now name the member and print the node, the way `XmlUnknownAttribute` does.

Closes #1719


morphqdd:
@yegor256, please take a look when you have a moment.

yegor256:
Merged, thanks!

---

